PR #11, opened 2019-06-25
[WIP] Idea for Integration Test

hayesgm:
Just a sketch/ WIP of thinking about integration tests (see integration.test.js not integration.sh). The idea is to run a few docker containers in a jest test and test that everything works that way. This can then be part of our standard test suite (though it'll be slow, so it might be skipped by default).

Thoughts on this direction?

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,6 @@
 .build
 .circleci
 .git
+poster
+sdk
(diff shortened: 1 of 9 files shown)


jflatow:
Closing since we merged integrations based on this already

PR #13, opened 2019-06-26
try timestamps without any Date hijinx to make test pass ci

coburncoburn:
reporter integratino test was flaky / brokne on ci, I swiched it to use a hard coded timestamp rather than Date module, seems better.

Also was not returning timestamp in payload of sample implemntation


sdk/javascript/tests/express_endpoint_test.ts
```diff
@@ -3,7 +3,7 @@ import {endpoint} from '../src/express_endpoint';
 
 test('integration test', async () => {
   const privateKey = '0x177ee777e72b8c042e05ef41d1db0f17f1fcb0e8150b37cfad6993e4373bdf10';
-  const timestamp = +new Date(2019, 6, 20);
+  const timestamp = 1563606000000;
```
jflatow: Doesn't really matter for this case but maybe drop 3 0s to save future confusion?

hayesgm:
yay, 3/3 passing!

PR #16, opened 2019-06-28
Trap reverts in poster, uppercase symbols, and 6 decimals w/ some minor other fixes

jflatow:


sdk/javascript/src/reporter.ts
```diff
@@ -40,7 +45,12 @@ export function fancyParameterEncoder(paramType: string): [string, (any) => any]
   // Decimals are encoded as uints with 18 decimals of precision on-chain.
   if (paramType === 'decimal') {
     actualParamType = 'uint256';
-    actualParamEnc = (x) => web3.utils.toBN('1000000000000000000').muln(x).toString();
+    actualParamEnc = (x) => web3.utils.toBN(1e6).muln(x).toString();
+  }
+
+  if (paramType == 'symbol') {
+    actualParamType = 'string';
+    actualParamEnc = (x) => x.toUpperCase();
```
hayesgm: upppppppppercase??????

jflatow: You mean which do we choose? Rob suggested upper

hayesgm: yeah just wondering why since upper case letter are so DiFfiCulT to write.

jflatow: I guess it makes sense since tickers are usually upper 🤷‍♂

hayesgm: hm, okay, do it!

PR #19, opened 2019-07-08
Add a symbols filter to the endpoint

jflatow:


sdk/javascript/src/express_endpoint.ts
```diff
@@ -1,24 +1,56 @@
...
+  return str.toUpperCase();
+}
+
+export const Filters = {
+  symbols: (pairs: Pairs, query: Query): Pairs => {
+    if (query.symbols) {
+      const symbols = new Set(flatStringList(query.symbols || []).map(upper));
+      if (Array.isArray(pairs))
+        return pairs.filter(([k, _]) => symbols.has(upper(k)));
+      return Object.entries(pairs).reduce((a, [k, v]) => {
```
coburncoburn: kind of lost on what the pair param is

coburncoburn: like it's an object or an array?, also it is compared to the symbols filter?

jflatow: Pairs is either a list of key, values or an object (was/is allowed by encoder) - this is to preserve whatever the input form given was on the filtered output - on the one hand kind of sucks to have this code but otoh much nicer to match what was given IMO

PR #20, opened 2019-07-08
Sign each price pair individually so poster can throttle

jflatow:
If we do these in bulk, then the symbols need to get pushed all the way to the data contract which is a bit awkward and also potentially gassy since it needs to loop through the symbols list for each key. This is actually not too bad gas-wise and probably a better API. Its also a relatively small change to encoding so hopefully not much extra work for partners who've already started building.


jflatow:
Still working on some changes in reporter, it does kind of break encode/decode symmetry currently

tests/DelFiPriceTest.js
```diff
@@ -76,7 +74,7 @@ describe('DelFiPrice', () => {
         ['ETH', 255]
       ]
     ]);
-    expect(post2.gasUsed).toBeLessThan(320000);
+    expect(post2.gasUsed).toBeLessThan(340000);
```
coburncoburn: so additional 2k gas per asset 🤷‍♂

jflatow:
I kept the `encode(kind, timestamp, pairs)` since its slightly more convenient and decoding is really only for sanity checking anyway

PR #21, opened 2019-07-09
Few items to save spiralling gas costs and allow poster optimization

jflatow:
- pack uint64 instead of uint256 so we use only a single storage slot
 - add back the symbols param to avoid building the list on-chain
  also gives some flexibility to poster
 - add more extensive gas testing

Re: other optimizations:
 - throttling by poster is best supported by having messages signed
 individually
 - view could receive median as a param and verify instead of
 calculating
  this would save the sort, but the sort is cheap compared to the cost
  of reading all the prices from the data contract

### DIFF
--- a/contracts/DelFiPrice.sol
+++ b/contracts/DelFiPrice.sol
@@ -12,7 +12,7 @@ contract DelFiPrice is OpenOracleView {
     /**
      * @notice The mapping of medianized prices per symbol
      */
-    mapping(string => uint) public prices;
+    mapping(string => uint64) public prices;
 
     constructor(OpenOraclePriceData data_, address[] memory sources_) public OpenOracleView(data_, sources_) {}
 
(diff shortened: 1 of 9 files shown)


contracts/DelFiPrice.sol
```diff
@@ -22,33 +22,16 @@ contract DelFiPrice is OpenOracleView {
      * @param messages The messages to post to the oracle
      * @param signatures The signatures for the corresponding messages
      */
-    function postPrices(bytes[] calldata messages, bytes[] calldata signatures) external {
+    function postPrices(bytes[] calldata messages, bytes[] calldata signatures, string[] calldata symbols) external {
```
coburncoburn: I hesitate to ask once again, but with the storage layout improvements, did we still think that it was important to have this to save gas, or was it that building the set is just kind of gross?

jflatow: It was costly

hayesgm: I *think* it's actually so you can take in a list of 10 reported prices but only pay to post 3 of them.

coburncoburn: to only post 3 of them it would be better to only post the 3 you like ( since they are individually signed )

contracts/DelFiPrice.sol
```diff
@@ -22,37 +27,22 @@ contract DelFiPrice is OpenOracleView {
...
-                if (keccak256(abi.encodePacked(symbol)) == keccak256(abi.encodePacked(symbols[j]))) {
-                    found = true;
-                    break;
-                }
-            }
-            if (!found) {
-                symbols[i] = symbol;
-                numSymbols += 1;
-            }
+            OpenOraclePriceData(address(data)).put(messages[i], signatures[i]);
```
coburncoburn: I notice that OpenOraclePriceData#put still returns the key / symbol. Do you want to keep that since it isn't being used anymore?

contracts/DelFiPrice.sol
jflatow: I feel like we should be ok making the post signature more complex to assist on-chain. I think both are nice features: to be able to just put data without recalculating median sometimes, or recalculate without putting data, and also to save some gas / code surface area (there was at least one bug in the prior impl). Not sure what the real downside is?

contracts/DelFiPrice.sol
jflatow: Ya intentional - I think its still useful in general or else callers need to know how to decode to extract the key that was written.

contracts/DelFiPrice.sol
coburncoburn: less solidity I think wins the day, but in general it's just kind of a code smell to me that we would ever need it in the first place, I suppose in the end it's just indicating the limitations of solidity, but wanted to make sure it wasn't indicating an issue in what we were doing.

coburncoburn: like, put prices + poke median would probably be the more clear design if we didn't have the assumption that we want to poke median almost every transaction that puts prices

jflatow: Yeah I think 2 functions would be less conflated but probably less convenient too, that said I don't think I have the same aversion to the 3rd arg

PR #23, opened 2019-08-05
Coburn/debug gas limit

coburncoburn:
nevermind the fact that this took 7 deployments


poster/src/index.ts
```diff
@@ -38,7 +39,8 @@ async function run() {
   }
 
   try {
-    await main(argv.sources, argv.posterKey, argv.viewAddress, argv.viewFunction, web3);
+    web3.eth.transactionConfirmationBlocks = 10
```
hayesgm: Just curious-- this is going to break dev right, since you only get one block confirmation on ganache?

coburncoburn: good point, did no mean for this to overwrite that dev change.

dustpan28645:
Be honest was right is right what's wrong is wrong !!!!💯
**tips**Are**Welcome**
****in**filecoin* f1j4oqo6bf5ru6fnpf2qqmyj3ecejtz7jjelgntva

PR #32, opened 2019-12-11
Coburn/fix integration test

coburncoburn:


poster/src/index.ts
```diff
@@ -22,31 +22,19 @@ async function run() {
...
-        }
-      }
-    }
+  } else {
+    web3.eth.transactionConfirmationBlocks = 10;
   }
 
   try {
     await main(argv.sources, argv.poster_key, argv.view_address, argv.view_function, argv.gas_limit, web3);
+    console.log("main compleed")
```
hayesgm: compleeeeeeted

poster/src/index.ts
```diff
@@ -22,31 +22,19 @@ async function run() {
...
   } catch (e) {
     console.error(`Poster failed to run`, e);
   }
 }
 
-run()
+run().then((x) => console.log(x));
```
jflatow: Still needed?

tests/IntegrationTest.js
```diff
@@ -7,19 +7,19 @@ const root = path.join(__dirname, '..');
 
 describe('Integration', () => {
   it('deploys the contracts, starts reporters and posts the right prices', async () => {
-  await compose.upAll({cwd: root, log: true});
-  try {
-    await new Promise(ok => setTimeout(ok, 3000));
+    try {
+      await compose.upOne(["poster"], {cwd: root, log: true});
```
jflatow: Was this for debugging?

coburncoburn: kind of works better this way since the poster has all the required dependencies

poster/src/index.ts
coburncoburn: double checked, it is indeed required,

PR #33, opened 2020-01-22
Upgrade to Newest Version of Saddle

hayesgm:
This patch upgrades to the newest version of saddle and fixes most of our docker items to be more resiliant to any number of changes.


docker-compose.yml
```diff
@@ -35,13 +39,19 @@ services:
...
     build: poster
-    command: npm run start --
+    command: "sh -c '
+      apk add jq;
+      while [ ! -f /build/development.json ]; do echo waiting for /build/development.json; sleep 1; done;
+      npm run start --
       --poster_key 0x4f3edf983ac636a65a842ce7c78d9aa706d3b113bce9c46f30d7d21715b23b1d
       --sources http://reporter-1:3000/prices.json,http://reporter-2:3000/prices.json
-      --view_address 0x5b1869d9a4c187f2eaa108f3062412ecf0526b24
+      --view_address $$(cat /build/development.json | jq -r .DelFiPrice)
```
jflatow: Way less fragile, but also some cognitive overhead

hayesgm: TBH, the hard-coded number was pretty confusing-- since it's mixed in with hard-coded reporter private keys/addresses

jflatow: Yeah I think this is a win, it would be really nice if docker respected the depends_on and it didn't have to do the wait though

coburncoburn:
so the docker composition is all sitting over there together, and only interface for it is on the web3 calls?

PR #34, opened 2020-01-31
Max/view🔭

maxwolff:
* adds an anchor price slot. we confirm median is close to the anchor before updating the median
* fix median to work with even numbers
* check timestamp before saving checkpoint
* refactor tests


contracts/DelFiPrice.sol
```diff
@@ -30,19 +39,23 @@ contract DelFiPrice is OpenOracleView {
...
+            uint64 medianPrice = medianPrice(symbol, sources);
+            uint64 anchorPrice = OpenOraclePriceData(address(data)).getPrice(anchor, symbol);
+            uint anchorRatioMantissa = medianPrice.mul(1e18).div(anchorPrice);
 
-            // Calculate the median price, write to storage, and emit an event
-            uint64 price = medianPrice(symbol, sources);
-            prices[symbol] = price;
-            emit Price(symbol, price);
+            // Only update the view's price if the median of the sources is within a bound
+            if (anchorRatioMantissa < (1e18).add(anchorToleranceMantissa) && anchorRatioMantissa > (1e18).sub(anchorToleranceMantissa)) {
```
maxwolff: minimum correct implementatation here. room for some gas optimization..

maxwolff: also worth thought about how / where we use safemath.

contracts/DelFiPrice.sol
```diff
@@ -62,7 +75,16 @@ contract DelFiPrice is OpenOracleView {
         }
 
         uint64[] memory sortedPrices = sort(postedPrices);
-        return sortedPrices[N / 2];
+
+        // if N is even, get the left and right medians and average them
+        if (N % 2 == 0) {
+            uint64 left = sortedPrices[(N / 2) - 1];
+            uint64 right = sortedPrices[N / 2];
+            return (left + right) / 2;
```
maxwolff: needs overflow check. safemath and then cast? or inline?

contracts/DelFiPrice.sol
```diff
@@ -3,23 +3,32 @@ pragma experimental ABIEncoderV2;
...
 import "./OpenOraclePriceData.sol";
 import "./OpenOracleView.sol";
+import "./SafeMath.sol";
 
 /**
  * @notice The DelFi Price Feed View
  * @author Compound Labs, Inc.
  */
 contract DelFiPrice is OpenOracleView {
+    using SafeMath for *;
```
coburncoburn: might as well using "SafeMath for uint256" here since it will revert for everything else. Interesting that this is an option though

contracts/DelFiPrice.sol
coburncoburn: I think left is  n/2 and right n/2 +1 

n = 4
left = 4/2 = 2
right = 4/2 + 1 = 3

contracts/DelFiPrice.sol
coburncoburn: oooh it upcasts things? interesting.

contracts/DelFiPrice.sol
maxwolff: n is length. so with a zero-indexed array with 4 objects, you'd actually want indexes 1 and 2.

contracts/DelFiPrice.sol
```diff
@@ -9,17 +9,28 @@ import "./OpenOracleView.sol";
...
 contract DelFiPrice is OpenOracleView {
+
     /**
      * @notice The event emitted when a price is written to storage
      */
     event Price(string symbol, uint64 price);
 
+    address anchor;
+    uint256 upperBoundAnchorRatio;
+    uint256 lowerBoundAnchorRatio;
```
jflatow: Prob want some natspecs on these, maybe as well change to the `///` form?

contracts/DelFiPrice.sol
```diff
@@ -9,17 +9,28 @@ import "./OpenOracleView.sol";
...
+
     /**
      * @notice The mapping of medianized prices per symbol
      */
     mapping(string => uint64) public prices;
 
-    constructor(OpenOraclePriceData data_, address[] memory sources_) public OpenOracleView(data_, sources_) {}
+    constructor(OpenOraclePriceData data_, address[] memory sources_, address anchor_, uint anchorToleranceMantissa_) public OpenOracleView(data_, sources_) {
+        anchor = anchor_;
+        //anchorMantissa of 10e17 means we tolerate a 10% difference between anchor and median
```
jflatow: 10e17 written 1e18?

maxwolff: typooo

contracts/DelFiPrice.sol
maxwolff: done

contracts/DelFiPrice.sol
```diff
@@ -9,40 +9,60 @@ import "./OpenOracleView.sol";
...
 
-            // Calculate the median price, write to storage, and emit an event
-            uint64 price = medianPrice(symbol, sources);
-            prices[symbol] = price;
-            emit Price(symbol, price);
+            // Only update the view's price if the median of the sources is within a bound
+            if (anchorRatioMantissa <= upperBoundAnchorRatio && anchorRatioMantissa >= lowerBoundAnchorRatio) {
+                prices[symbol] = medianPrice;
+                emit Price(symbol, medianPrice);
+            }
```
jflatow: So we quietly omit out of bounds prices...this raises the question about how we know when its happening?

contracts/DelFiPrice.sol
```diff
@@ -62,7 +82,17 @@ contract DelFiPrice is OpenOracleView {
...
 
         uint64[] memory sortedPrices = sort(postedPrices);
-        return sortedPrices[N / 2];
+
+        // if N is even, get the left and right medians and average them
+        if (N % 2 == 0) {
+            uint64 left = sortedPrices[(N / 2) - 1];
+            uint64 right = sortedPrices[N / 2];
+            require((left + right) >= left, "DelfiPrice::MedianPrice price addition overflow");
+            return (left + right) / 2;
```
jflatow: Alternatively we could use an intermediate uint or uint128 for left + right, since we know (left + right) / 2 will fit in uint64. Saves a tiny bit of gas but also might look cleaner 🤷‍♂

tests/OpenOracleDataTest.js
```diff
@@ -9,12 +10,21 @@ const {
...
+  let anchorAddress = accounts[1];
+
+  beforeEach(async done => {
+    oracleData = await deploy('OpenOracleData', []);
+    priceData = await deploy('OpenOraclePriceData', []);
+    delfi = await deploy('DelFiPrice', [priceData._address, [account], anchorAddress, numToHex(1e17)]);
+    done();
+  });
+
+  // TODO: refactor this ;/
```
jflatow: Yeah I really with describe could be async too

contracts/DelFiPrice.sol
maxwolff: well, we'd see `Write` events in the Data contract, and no corresponding `Price` event for the same tx. think we should add an event here?

coburncoburn: we have capped price posted in current set up and it is very useful, +1 on adding an event for this

contracts/DelFiPrice.sol
```diff
@@ -9,40 +9,60 @@ import "./OpenOracleView.sol";
...
+    /**
+     * @param data_ Address of the Oracle Data contract
+     * @param sources_ The reporter addresses whose prices will be used to calculate the median
+     * @param anchor_ The reporter address whose prices checked against the median for safety
+     * @param anchorToleranceMantissa_ The tolerance allowed between the anchor and median. A tolerance of 1e17 means a new median that is 10% off from the anchor will still be saved
+     */
+    constructor(OpenOraclePriceData data_, address[] memory sources_, address anchor_, uint anchorToleranceMantissa_) public OpenOracleView(data_, sources_) {
+        anchor = anchor_;
+        require(anchorToleranceMantissa_ < 1e18, "Anchor Tolerance is too high");
+        upperBoundAnchorRatio = 1e18 + anchorToleranceMantissa_;
```
hayesgm: Just to verify, this is basically meant to be `100% + 10%` and `100% - 10%`, right?

contracts/DelFiPrice.sol
```diff
@@ -9,40 +9,60 @@ import "./OpenOracleView.sol";
...
         for (uint i = 0; i < messages.length; i++) {
             OpenOraclePriceData(address(data)).put(messages[i], signatures[i]);
         }
 
         // Recalculate the asset prices for the symbols to update
         for (uint i = 0; i < symbols.length; i++) {
             string memory symbol = symbols[i];
+            uint64 medianPrice = medianPrice(symbol, sources);
+            uint64 anchorPrice = OpenOraclePriceData(address(data)).getPrice(anchor, symbol);
+            uint256 anchorRatioMantissa = uint256(medianPrice) * 1e18 / anchorPrice;
```
hayesgm: Do we test/handle when anchorPrice is 0?

contracts/DelFiPrice.sol
```diff
@@ -9,40 +9,60 @@ import "./OpenOracleView.sol";
...
 
-            // Calculate the median price, write to storage, and emit an event
-            uint64 price = medianPrice(symbol, sources);
-            prices[symbol] = price;
-            emit Price(symbol, price);
+            // Only update the view's price if the median of the sources is within a bound
+            if (anchorRatioMantissa <= upperBoundAnchorRatio && anchorRatioMantissa >= lowerBoundAnchorRatio) {
```
hayesgm: Should we emit an event when the price is outside the bound?

contracts/DelFiPrice.sol
```diff
@@ -62,7 +82,17 @@ contract DelFiPrice is OpenOracleView {
...
-        return sortedPrices[N / 2];
+
+        // if N is even, get the left and right medians and average them
+        if (N % 2 == 0) {
+            uint64 left = sortedPrices[(N / 2) - 1];
+            uint64 right = sortedPrices[N / 2];
+            require((left + right) >= left, "DelfiPrice::MedianPrice price addition overflow");
+            return (left + right) / 2;
+        } else {
+        // if N is odd, just return the median
```
hayesgm: spacing

contracts/DelFiPrice.sol
maxwolff: yes

contracts/DelFiPrice.sol
maxwolff: good point, working on it

contracts/DelFiPrice.sol
maxwolff: yes, others also agreed above. working on it

tests/DelFiPriceTest.js
```diff
@@ -58,8 +58,7 @@ async function setup(N) {
         ),
         signers[i].privateKey
       );
-      for (let { message, signature, signatory } of signed) {
-        // expect(signatory).toEqual(signers[i].address);
```
maxwolff: we test this elsewhere

contracts/DelFiPrice.sol
jflatow: Bit of a nit but I still like the convention for XXe16 for % mantissas, we always get confused about scales and we started to do that in other contracts which I think makes it relatively clear

contracts/DelFiPrice.sol
```diff
@@ -9,40 +9,71 @@ import "./OpenOracleView.sol";
...
+                uint256 anchorRatioMantissa = uint256(medianPrice) * 1e18 / anchorPrice;
+                // Only update the view's price if the median of the sources is within a bound, and it is a new median
+                if (anchorRatioMantissa <= upperBoundAnchorRatio && anchorRatioMantissa >= lowerBoundAnchorRatio) {
+                    // only update and emit event if the median price is new, otherwise do nothing
+                    if (prices[symbol] != medianPrice) {
+                        prices[symbol] = medianPrice;
+                        emit PriceUpdated(symbol, medianPrice);
+                    }
+                } else {
+                    emit PriceGuarded(symbol, medianPrice, anchorPrice);
```
jflatow: Great name 🛡

contracts/DelFiPrice.sol
```diff
@@ -62,7 +93,18 @@ contract DelFiPrice is OpenOracleView {
...
 
         uint64[] memory sortedPrices = sort(postedPrices);
-        return sortedPrices[N / 2];
+
+        // if N is even, get the left and right medians and average them
+        if (N % 2 == 0) {
+            uint64 left = sortedPrices[(N / 2) - 1];
+            uint64 right = sortedPrices[N / 2];
+            uint128 sum = left + right;
+            require(sum >= left, "DelfiPrice::MedianPrice price addition overflow");
```
jflatow: 🔪 ?

contracts/OpenOracleView.sol
```diff
@@ -26,6 +26,7 @@ contract OpenOracleView {
      * @param sources_ The list of source addresses to include in the aggregate value
      */
     constructor(OpenOracleData data_, address[] memory sources_) public {
+        require(sources_.length > 0, "Must initialize with sources");
```
jflatow: 🛡

contracts/OpenOraclePriceData.sol
```diff
@@ -43,9 +41,11 @@ contract OpenOraclePriceData is OpenOracleData {
 
         // Only update if newer than stored, according to source
         Datum storage prior = data[source][key];
-        if (prior.timestamp < timestamp) {
+        if (prior.timestamp < timestamp && timestamp < block.timestamp + 1 hours) {
```
jflatow: I wonder if 60 minutes > 1 hours

contracts/DelFiPrice.sol
maxwolff: 🔪

tests/contracts/Test.sol
```diff
@@ -0,0 +1,9 @@
+pragma solidity ^0.5.12;
+pragma experimental ABIEncoderV2;
+
+contract Test {
+
+    function testOverflow() public pure {
+        require(uint64(uint128(uint64(-1) + uint64(-1))/2) > uint64(10000), 'overflows');
```
jflatow: Still think we might as well assert it exactly equals the correct result but 🤷‍♂

contracts/DelFiPrice.sol
hayesgm: Event was added.

contracts/DelFiPrice.sol
```diff
@@ -62,7 +93,16 @@ contract DelFiPrice is OpenOracleView {
         }
 
         uint64[] memory sortedPrices = sort(postedPrices);
-        return sortedPrices[N / 2];
+        // if N is even, get the left and right medians and average them
+        if (N % 2 == 0) {
+            uint64 left = sortedPrices[(N / 2) - 1];
```
hayesgm: Is this right? My intuition is that is about be `N/2` and `N/2+1` not `N/2` and `N/2-1`, but that's just intuition.

contracts/OpenOraclePriceData.sol
```diff
@@ -43,9 +41,11 @@ contract OpenOraclePriceData is OpenOracleData {
 
         // Only update if newer than stored, according to source
         Datum storage prior = data[source][key];
-        if (prior.timestamp < timestamp) {
+        if (prior.timestamp < timestamp && timestamp < block.timestamp + 60 minutes) {
```
hayesgm: can we change the first clause here to `timestamp > prior.timestamp` that I think reads more obviously

contracts/DelFiPrice.sol
maxwolff: the array is zero-indexed, and N refers to lenght. so for example a 4 item array has indexes 0,1,2,3 and N of 4. We want indices 1 and 2, so N/2 and N/2 -1 is correct. 

might be worth adding a comment since it has come up twice now.

contracts/OpenOraclePriceData.sol
maxwolff: yes, done

PR #35, opened 2020-02-04
add timestamp checks

maxwolff:
* check the signed timestamp on price messages is not in the future, not before the prior timestamp, and in the last 10 minutes
* refactored some tests

### DIFF
--- a/contracts/OpenOraclePriceData.sol
+++ b/contracts/OpenOraclePriceData.sol
@@ -8,14 +8,12 @@ import "./OpenOracleData.sol";
  * @author Compound Labs, Inc.
  */
 contract OpenOraclePriceData is OpenOracleData {
-    /**
-     * @notice The event emitted when a source writes to its storage
-     */
+    ///@notice The event emitted when a source writes to its storage
     event Write(address indexed source, string key, uint64 timestamp, uint64 value);
+    ///@notice The event emitted when the timestamp on a price is invalid and it is not written to storage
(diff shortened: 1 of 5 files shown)


contracts/OpenOraclePriceData.sol
```diff
@@ -43,9 +41,11 @@ contract OpenOraclePriceData is OpenOracleData {
 
         // Only update if newer than stored, according to source
         Datum storage prior = data[source][key];
-        if (prior.timestamp < timestamp) {
+        if (prior.timestamp < timestamp && timestamp < block.timestamp && block.timestamp - 10 minutes < timestamp) {
```
maxwolff: check the message was signed in the last 10 minutes. totally random guess at the value we want, up for discussion

jflatow: This might be a little stringent, I have mixed feelings about relying too much on block.timestamp since it slightly creates a new risk vector. What about just timestamp < block.timestamp + 10 minutes?

maxwolff: yeah, a buffer for future timestamps makes sense. you dont think we should prevent posting stale prices? what about if the limit was 1 hour

jflatow: I'm trying to think why we should ever reject a price that's fresher than the one on-chain

maxwolff: i actually kind of agree.. @hayesgm ?

PR #41, opened 2020-04-11
add rotation helper

coburncoburn:


sdk/javascript/src/reporter.ts
```diff
@@ -74,6 +74,10 @@ export function encode(kind: string, timestamp: number, pairs: [any, any][] | ob
   });
 }
 
+export function encodeRotationMessage(rotationTarget: string) : string {
+  return web3.eth.abi.encodeParameters(['string', 'address'], ['rotate', rotationTarget]);
+}
+
```
jflatow, line 80: lgtm...though still a little weird to me still that its sort of a generic message but actually its a rotation message

coburncoburn: so compared to the prices one, it has kind, timestamp, values? u think we should do kind, timestamp, args? seems kind of okay actually..
`
"rotatation", time, [rotationTarget
`
such that all messages come that way?

jflatow: Yeah, that works - timestamp seems like a good idea actually. Or alternatively we never pretend its some kind of general self-describing message...just in between seems suboptimal

coburncoburn: what is the inbetween though? you need some sort of tag. seems like the current state is the most concrete you can get

jflatow: Yeah I guess in this repo its actually quite concrete, but the way its specified makes it seem like it can have an action which is not `"rotate"`, but that if it did, the arg would still be called `rotateTarget`

PR #45, opened 2020-04-15
PR for discussion, add ctokens to symbols mapping

antonina-cherednichenko:
Questions for discussion:

1. Does this all makes sense? 
2. 8 additional parameters for constructor is way too many, maybe I can pass a struct since we already using ABIEncoderV2. I need to do more research on that. 
3. uint256 (from oracle proxy) --> uint64? Can it be potentially dangerous?
4. Rename methods? 
5. What to do with SAI?
6. Interface for protocol oracle proxy with getUnderlying price, address vs cToken parameter? @coburncoburn @jflatow


contracts/DelFiPrice.sol
```diff
@@ -27,17 +32,58 @@ contract DelFiPrice is OpenOracleView {
...
+                address[] memory sources_,
+                address anchor_,
+                uint anchorToleranceMantissa_,
+                address cEthAddress_,
+                address cUsdcAddress_,
+                address cDaiAddress_,
+                // address cRepAddress_,
+                // address cWbtcAddress_, 
+                // address cBatAddress_, 
+                address cZrxAddress_) public OpenOracleView(data_, sources_) {
```
jflatow, line 75: We can use a list if we need to (like sources), its just hard to do ordering - slightly surprised this doesn't work though since I think we have other constructors with as many args as this

contracts/DelFiPrice.sol
```diff
@@ -105,6 +152,23 @@ contract DelFiPrice is OpenOracleView {
...
+     */
+    function getOracleKey(string memory symbol) public view returns (address cToken) {
+        if (keccak256(abi.encodePacked(symbol)) == keccak256(abi.encodePacked("ETH"))) return cEthAddress;
+        if (keccak256(abi.encodePacked(symbol)) == keccak256(abi.encodePacked("USDC"))) return cUsdcAddress;
+        if (keccak256(abi.encodePacked(symbol)) == keccak256(abi.encodePacked("DAI"))) return cDaiAddress;
+        if (keccak256(abi.encodePacked(symbol)) == keccak256(abi.encodePacked("REP"))) return cRepAddress;
+        if (keccak256(abi.encodePacked(symbol)) == keccak256(abi.encodePacked("BTC"))) return cWbtcAddress;
+        if (keccak256(abi.encodePacked(symbol)) == keccak256(abi.encodePacked("BAT"))) return cBatAddress;
+        if (keccak256(abi.encodePacked(symbol)) == keccak256(abi.encodePacked("ZRX"))) return cZrxAddress;
+        return address(0);
```
jflatow, line 169: Should this revert?

contracts/DelFiPrice.sol
antonina-cherednichenko: 3 parameters too many :(  

```
Got this Error: Command failed: solc --combined-json bin,abi,metadata --metadata-literal --optimize  contracts/*.sol
Exception during compilation: /solidity/libyul/backends/evm/AsmCodeGen.cpp(211): Throw in function static void solidity::yul::CodeGenerator::assemble(const solidity::yul::Block&, solidity::yul::AsmAnalysisInfo&, solidity::evmasm::Assembly&, solidity::langutil::EVMVersion, const solidity::yul::ExternalIdentifierAccess&, bool, bool)
Dynamic exception type: boost::wrapexcept<solidity::yul::YulAssertion>
std::exception::what: Stack too deep when compiling inline assembly: Variable headStart is 2 slot(s) too deep inside the stack.
[solidity::util::tag_comment*] = Stack too deep when compiling inline assembly: Variable headStart is 2 slot(s) too deep inside the stack.
```

antonina-cherednichenko: Tried to compile with --optimize-yul flag, but no luck, the same error

contracts/DelFiPrice.sol
antonina-cherednichenko: yes, makes sense.

contracts/DelFiPrice.sol
```diff
@@ -27,17 +32,58 @@ contract DelFiPrice is OpenOracleView {
     /// @notice The mapping of medianized prices per symbol
     mapping(string => uint64) public prices;
 
+    /// @notice Address of the cEther contract
+    address public cEthAddress;
```
coburncoburn, line 36: we want these as immutable 
https://solidity.readthedocs.io/en/v0.6.6/contracts.html#constant-and-immutable-state-variables

contracts/DelFiPrice.sol
```diff
@@ -105,6 +152,23 @@ contract DelFiPrice is OpenOracleView {
...
     }
 
+
+    /**
+     * @notice Returns the cToken address for symbol
+     * @param symbol The symbol to map to cToken address
+     * @return cToken The cToken address for the given symbol
+     */
+    function getOracleKey(string memory symbol) public view returns (address cToken) {
+        if (keccak256(abi.encodePacked(symbol)) == keccak256(abi.encodePacked("ETH"))) return cEthAddress;
```
coburncoburn, line 162: we could keccack the arg symbol once at the beginning, and put the symbols for comparison in contants to make this more palatable

contracts/DelFiPrice.sol
coburncoburn: struct arg could come out nice, or a 2d array [[ctokens], [symbols]] or [[ctoken, symbol], [ctoken, symbol]] are a couple of other ways we may be able to slip out of this one

contracts/DelFiPrice.sol
coburncoburn: eg constant symbolDai = keccak256(abi.encodePacked("DAI")) etc etc

contracts/DelFiPrice.sol
antonina-cherednichenko: makes sense, I will try both approaches.

contracts/DelFiPrice.sol
```diff
@@ -105,6 +152,24 @@ contract DelFiPrice is OpenOracleView {
...
     }
 
+
+    /**
+     * @notice Returns the cToken address for symbol
+     * @param symbol The symbol to map to cToken address
+     * @return cToken The cToken address for the given symbol
+     */
+    function getCTokenAddress(string memory symbol) public view returns (address cToken) {
+        bytes32 symbolHash = keccak256(abi.encodePacked(symbol));
```
jflatow, line 162: Oh not too bad at all now

contracts/DelFiPrice.sol
```diff
@@ -15,6 +20,17 @@ contract DelFiPrice is OpenOracleView {
...
+    struct CTokens {
+        address cEthAddress;
+        address cUsdcAddress;
+        address cDaiAddress;
+        address cRepAddress;
+        address cWbtcAddress; 
+        address cBatAddress; 
+        address cZrxAddress;
+    }
+
```
jflatow, line 33: Nice 🍠

contracts/DelFiPrice.sol
```diff
@@ -27,17 +43,47 @@ contract DelFiPrice is OpenOracleView {
...
      * @param anchor_ The reporter address whose prices checked against the median for safety
      * @param anchorToleranceMantissa_ The tolerance allowed between the anchor and median. A tolerance of 10e16 means a new median that is 10% off from the anchor will still be saved
+     * @param tokens_ The CTokens struct that contains addresses for CToken contracts
      */
-    constructor(OpenOraclePriceData data_, address[] memory sources_, address anchor_, uint anchorToleranceMantissa_) public OpenOracleView(data_, sources_) {
+    constructor(OpenOraclePriceData data_, 
+                address[] memory sources_,
+                address anchor_,
+                uint anchorToleranceMantissa_,
+                CTokens memory tokens_) public OpenOracleView(data_, sources_) {
```
coburncoburn, line 81: struct constructor arg is very cool

contracts/DelFiPrice.sol
```diff
@@ -27,17 +43,47 @@ contract DelFiPrice is OpenOracleView {
...
         anchor = anchor_;
         require(anchorToleranceMantissa_ < 100e16, "Anchor Tolerance is too high");
         upperBoundAnchorRatio = 100e16 + anchorToleranceMantissa_;
         lowerBoundAnchorRatio = 100e16 - anchorToleranceMantissa_;
+        tokens = tokens_;
```
coburncoburn, line 86: we probably want to destructure the tokens into immutable state variables to save all that gas in the lookup function. Would we awesome if immutable structs were a thing

antonina-cherednichenko: cool, so I will change that for this variant with structure. And then we can look at how 2 dimensional array variant looks like as well.

contracts/DelFiPrice.sol
```diff
@@ -105,6 +176,24 @@ contract DelFiPrice is OpenOracleView {
...
 
+
+    /**
+     * @notice Returns the cToken address for symbol
+     * @param symbol The symbol to map to cToken address
+     * @return cToken The cToken address for the given symbol
+     */
+    function getCTokenAddress(string memory symbol) public view returns (address cToken) {
+        bytes32 symbolHash = keccak256(abi.encodePacked(symbol));
+        if (symbolHash == symbolEth) return cEthAddress;
```
hayesgm, line 187: boy, I miss case/pattern match statements!

contracts/DelFiPrice.sol
```diff
@@ -58,7 +128,8 @@ contract DelFiPrice is OpenOracleView {
         for (uint i = 0; i < symbols.length; i++) {
             string memory symbol = symbols[i];
             uint64 medianPrice = medianPrice(symbol, sources);
-            uint64 anchorPrice = OpenOraclePriceData(address(data)).getPrice(anchor, symbol);
+            // uint64 anchorPrice = OpenOraclePriceData(address(data)).getPrice(anchor, symbol);
+            uint64 anchorPrice = uint64(AnchorPriceOracle(address(anchor)).getUnderlyingPrice(getCTokenAddress(symbol)));
```
hayesgm, line 132: Maybe break this into multiple steps? It's doing.. a lot for one 1 LoC

contracts/DelFiPrice.sol
antonina-cherednichenko: I added a second PR that gets an array of pairs in constructor. The final result is in a way much simpler than this original Pr.

contracts/DelFiPrice.sol
antonina-cherednichenko: yes, we can easily add it here.

tests/DelFiPriceTest.js
```diff
@@ -154,7 +187,7 @@ describe('DelFiPrice', () => {
     });
   });
 
-  describe('Deploy with an odd # of sources', () => {
+  describe.skip('Deploy with an odd # of sources', () => {
```
jflatow: Did you mean to commit this?

antonina-cherednichenko: I fixed that test and enabled it back later. I was trying to see if pipeline will pass

contracts/DelFiPrice.sol
antonina-cherednichenko: Moving one with this PR and immutable constants.

contracts/DelFiPrice.sol
antonina-cherednichenko: Actually no, it can not be added, not in solidity, in Yul.

contracts/DelFiPrice.sol
antonina-cherednichenko: I will do after merging Coburn's pr, so that we have the final version of this line before splitting it.

Dockerfile
```diff
@@ -16,6 +16,7 @@ RUN yarn install
 
 ENV PROVIDER PROVIDER
 ADD contracts contracts
+ADD tests tests
```
hayesgm, line 19: Do you actually run tests in Docker? I guess that's pretty cool for isolation

contracts/DelFiPrice.sol
hayesgm: Yeah, to put this another way: Solidity is not about cool. Solidity is about unlearning every programming technique you've learned in the last 10 years and writing things out like you were teaching a toddler. Make it direct and simple.

Dockerfile
antonina-cherednichenko: I had to add that for saddle config to find ProxyPriceOracle.sol in tests/contracts for mocking anchor oracle in tests.

antonina-cherednichenko:
Maybe let's add getUnderlyingPrice and IsPriceOracle to some interface that DelFiPrice will implement? To show that they these methods will be called by protocol proxy oracle? @coburncoburn @jflatow-compound @hayesgm

jflatow:
> Maybe let's add getUnderlyingPrice and IsPriceOracle to some interface that DelFiPrice will implement? To show that they these methods will be called by protocol proxy oracle? @coburncoburn @jflatow-compound @hayesgm

Good idea, we could actually implement this: https://github.com/compound-finance/compound-protocol/blob/master/contracts/PriceOracle.sol (well, just inline it like the other interface, but good to show that I think 👍)

contracts/DelFiPrice.sol
```diff
@@ -27,17 +45,85 @@ contract DelFiPrice is OpenOracleView {
...
+    bytes32 constant symbolWbtc = keccak256(abi.encodePacked("BTC"));
+
+    /// @notice The binary representation for 'BAT' symbol, used for string comparison
+    bytes32 constant symbolBat = keccak256(abi.encodePacked("BAT"));
+
+    /// @notice The binary representation for 'ZRX' symbol, used for string comparison
+    bytes32 constant symbolZrx = keccak256(abi.encodePacked("ZRX"));
+
+    /// @notice The binary representation for 'SAI' symbol, used for string comparison
+    bytes32 constant symbolSai = keccak256(abi.encodePacked("SAI"));
```
coburncoburn, line 70: we won't actually need SAI as a string, ( ever probably ) and actually will only need BTC and ETH until coinbase starts to add more assets

jflatow: Good point

antonina-cherednichenko:
Also maybe be can use bytes32 symbols for backward mapping and storing prices instead of having constants like "ETH" in getOracleKey.

contracts/DelFiPrice.sol
antonina-cherednichenko: deleted it in the last commit

coburncoburn:
superceded by #56

PR #46, opened 2020-04-15
PR for discussion, add ctokens to symbols mapping <--Show 1 more variant of token symbols to addresses implementation

antonina-cherednichenko:
Inspired by a way @coburncoburn suggested to pass arrays with symbols and CToken addresses in constructor. Unfortunately I cannot define an array with different types like `address` and `string` in Solidity. But I can pass an array with simple structs that contain address and string for each CToken pair. 

The final code here looks much simpler, but no immutable vars.


coburncoburn:
interesting, so since SAI is imminently locked, and USDC and USDT will be pegged to 1, we will have 6 prices being posted, and add about 800 * 6 = 4800 gas if posting all 6 at once, so probably NBD there

 OTOH, we need the reverse mapping for getUnderlyingPrice and not adding that gas to user actions is pretty desirable. I think we might as well save it while we can.

antonina-cherednichenko:
@coburncoburn makes sense, I will be working on variant 1 for now. The only thing I really like about this pr is an ability to easily add new market pairs.

contracts/DelFiPrice.sol
```diff
@@ -43,47 +37,8 @@ contract DelFiPrice is OpenOracleView {
...
-    /// @notice Address of the cWBTC contract
-    address public immutable cWbtcAddress;
-
-    /// @notice Address of the cBAT contract
-    address public immutable cBatAddress;
-
-    /// @notice Address of the cZRX contract
-    address public immutable cZrxAddress;
+    /// @notice The mapping of token symbols to CToken contracts addresses
+    mapping(string => address) tokens;
```
jflatow, line 41: Won't these be much gassier to read?

antonina-cherednichenko: yes, moving one with immutable constants.

PR #48, opened 2020-04-21
Replace constants representation with symbols

antonina-cherednichenko:
Replace usage of string currency constants with symbols everywhere.

### DIFF
--- a/contracts/DelFiPrice.sol
+++ b/contracts/DelFiPrice.sol
@@ -43,7 +43,7 @@ contract DelFiPrice is OpenOracleView {
     uint256 lowerBoundAnchorRatio;
 
     /// @notice The mapping of medianized prices per symbol
-    mapping(string => uint64) public prices;
+    mapping(bytes32 => uint64) public prices;
 
     /// @notice The binary representation for 'ETH' symbol , used for string comparison
     bytes32 constant symbolEth = keccak256(abi.encodePacked("ETH"));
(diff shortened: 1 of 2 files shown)


contracts/DelFiPrice.sol
```diff
@@ -238,16 +239,16 @@ contract DelFiPrice is OpenOracleView {
...
-        if (cToken == cUsdcAddress) return "USDC";
-        if (cToken == cDaiAddress) return "DAI";
-        if (cToken == cRepAddress) return "REP";
-        if (cToken == cWbtcAddress) return "BTC";
-        if (cToken == cBatAddress) return "BAT";
-        if (cToken == cZrxAddress) return "ZRX";
-        if (cToken == cSaiAddress) return "SAI";
-        if (cToken == cUsdtAddress) return "USDT";
+    function getOracleKey(address cToken) public view returns (bytes32) {
+        if (cToken == cEthAddress) return  symbolEth;
```
coburncoburn, line 243: I think I like this idea, what if we just went for using the cToken address for the view storage key , then we don't need this functino at all

antonina-cherednichenko:
@coburncoburn @jflatow @hayesgm Here is a way to consistently use crypto constant symbols that are defined at the beginning of contract, so that we don't use string constants in getOracleKey.
Drawback 'prices' mapping will use bytes32 instead of simple string symbol, but this probably can save gas from another point.

contracts/DelFiPrice.sol
antonina-cherednichenko: I like this idea. We can store prices as mapping from address => uint , when posting prices, map string to this address, and delete getOracleKey completely. 

@jflatow what do you think?

jflatow: So prices would be mapping cToken to price? I think since we have the string => price in the data contract its probably fine...but this was really a 'Rob constraint' originally so we need to ask him if he's ok with it

PR #49, opened 2020-04-22
Cob/decimals normalizing

coburncoburn:


contracts/DelFiPrice.sol
```diff
@@ -170,20 +175,66 @@ contract DelFiPrice is OpenOracleView {
...
+        address tokenAddress = getCTokenAddress(symbol);
+
+        if ( tokenAddress == cUsdcAddress || tokenAddress == cUsdtAddress )  {
+            // hard code to 1 dollar
+            return 1e6;
+        }
+
+        uint priceInEth = anchor.getUnderlyingPrice(tokenAddress);
+        uint additionalScale;
+        if ( tokenAddress == cWbtcAddress ){
```
hayesgm, line 194: Maybe we should have some kind of mapping that is "decimals" so we can just have a clean way of getting this? Here's a thought, we have an immutable `TokenData` struct that's returns from Toni's case statement below and passed around? It could have Oracle symbol, reporter list, decimals, etc, and just be hard-coded in the constructor. This could make life :oh so nice:

coburncoburn: I really wanted immutable structs to be a thing too, but they're not. (we could probably fake it better than this though )

coburncoburn: but setting in constants or a function, yes will take a look and see if it looks nicer

coburncoburn: oh you mean like an in memory struct? does that work? that would be next best thing to real `immutable` struct

contracts/DelFiPrice.sol
```diff
@@ -170,20 +175,66 @@ contract DelFiPrice is OpenOracleView {
      * @return true, this contract is meant to be used by Compound v2 PriceOracle interface.
      */
     function isPriceOracle() external pure returns (bool) {
-       return true;
+        return true;
+    }
+
+    // fetchs price in eth from proxy and converts to usd price using anchor usdc price
+    // anchor usdc price has 30 decimals, and anchor general price has 18 decimals,
```
antonina-cherednichenko, line 182: 30? or 36 decimals for usdc price?

coburncoburn:
superceded by #56

PR #51, opened 2020-04-22
Enable back IntegrationTest.js

antonina-cherednichenko:


tests/IntegrationTest.js
```diff
@@ -34,13 +33,13 @@ async function waitForLogs(serviceLogPairs) {
   if (complete) {
     return;
   } else {
-    await sleep(5000);
+    await sleep(10000);
```
coburncoburn, line 36: 💤
double the sleep 😴

antonina-cherednichenko:
The old way with using `contract`, but adding  `extra_build_files` seems to work
`const delfi = await contract.getContractAt(web3, 'DelFiPrice', {build_dir: '.dockerbuild_cp', trace: false, extra_build_files: []}, '0x5b1869D9A4C187F2EAa108f3062412ecf0526b24', {from: accounts[0]});`

PR #55, opened 2020-04-23
add coinbase auth

coburncoburn:


andrei-anisimov:
Nice, thanks @coburncoburn!

PR #56, opened 2020-04-23
Cob/clean anchor view

antonina-cherednichenko:
The version in which DelFiPrice is unchanged and all the changes with anchoring live inside AnchoredPriceView.


contracts/DelFiPrice.sol
```diff
@@ -123,4 +123,4 @@ contract DelFiPrice is OpenOracleView {
         }
         return array;
     }
-}
+}
```
antonina-cherednichenko, line 126: This no new line thing is annoying, any ideas how to disable that?

package.json
```diff
@@ -7,7 +7,7 @@
   "author": "Compound Labs, Inc.",
   "license": "MIT",
   "dependencies": {
-    "eth-saddle": "0.1.8",
+    "eth-saddle": "^0.1.17",
```
antonina-cherednichenko, line 10: "^0.1.17" vs "0.1.17"

Dockerfile
```diff
@@ -16,6 +16,7 @@ RUN yarn install
 
 ENV PROVIDER PROVIDER
 ADD contracts contracts
+ADD tests tests
 ADD saddle.config.js saddle.config.js
```
jflatow, line 20: Each of these ADDs actually adds a layer (slows down building, pushing), maybe we could copy them all on one line?

contracts/AnchoredPriceView.sol
```diff
@@ -0,0 +1,302 @@
...
+
+/**
+ * @notice Price feed conforming to Price Oracle Proxy interface and
+ * using a single open oracle reporter, anchored to and falling back to
+ * the Compound v2 oracle system.
+ * @author Compound Labs, Inc.
+ */
+contract AnchoredPriceView {
+    /// @notice standard amount for the Dollar
+    uint constant oneDollar = 1e6;
```
jflatow, line 18: Nice touch

contracts/AnchoredPriceView.sol
```diff
@@ -0,0 +1,302 @@
...
+
+    /// @notice The event emitted when the median price is updated
+    event PriceUpdated(string symbol, uint64 price);
+
+    /// @notice The event emitted when new prices are posted but the median price is not updated due to the anchor
+    event PriceGuarded(string symbol, uint64 median, uint64 anchor);
```
jflatow, line 24: Given what we found with gov wanting indices, do we want to index by symbol?

contracts/AnchoredPriceView.sol
```diff
@@ -0,0 +1,302 @@
...
+        revert("Unknown token address");
+    }
+
+    function mul(uint256 a, uint256 b) internal pure returns (uint256) {
+        if (a == 0) {
+            return 0;
+        }
+
+        uint256 c = a * b;
+        require(c / a == b, "SafeMath: multiplication overflow");
```
jflatow, line 297: Probably we can drop the "SafeMath:" and possibly pass in message like we sometimes do? Generic message is also fine if/since we have traces now

contracts/DelFiPrice.sol
jflatow: I think you'd have to add a newline

package.json
jflatow: 🤷

saddle.config.js
```diff
@@ -3,83 +3,110 @@ module.exports = {
...
+  contracts: process.env['SADDLE_CONTRACTS'] || "contracts/*.sol tests/contracts/*.sol",  // Glob to match contract files
+  tests: ['**/tests/*Test.js'],                         // Glob to match test files
   networks: {                                           // Define configuration for each network
     development: {
       providers: [                                      // How to load provider (processed in order)
-        {env: "PROVIDER"},                              // Try to load Http provider from `PROVIDER` env variable (e.g. env PROVIDER=http://...)
-        {http: "http://127.0.0.1:8545"}                 // Fallback to localhost provider
+        { env: "PROVIDER" },                              // Try to load Http provider from `PROVIDER` env variable (e.g. env PROVIDER=http://...)
+        { http: "http://127.0.0.1:8545" }                 // Fallback to localhost provider
       ],
```
jflatow, line 13: Another vscode user?

contracts/AnchoredPriceView.sol
```diff
@@ -0,0 +1,302 @@
...
+
+    /**
+     * @dev fetch price in eth from proxy and convert to usd price using anchor usdc price.
+     * @dev Anchor usdc price has 30 decimals, and anchor general price has 18 decimals, so multiplying by 1e18 and by 1e30 yields 1e6
+     */
+    function getAnchorPrice(address tokenAddress, uint256 usdcPrice) public view returns (uint64) {
+
+        if ( tokenAddress == cUsdcAddress || tokenAddress == cUsdtAddress )  {
+            // hard code to 1 dollar
+            return uint64(oneDollar);
```
jflatow, line 200: Oh actually this is pretty confusing - we pass in the `usdcPrice` but its not the `usdcPrice`?

coburncoburn: yes, usdcPrice ( in terms of Ether ) is used to convert other ether prices to dollars, but if actually wanting anchor price for usdc, we should always return 1 dollar. 

There is a little oddity here that we probably shouldn't even really be trying to post a price for usdc or usdt, but I thought we would want to guard against it in some form

Dockerfile
antonina-cherednichenko: I can look into that.

antonina-cherednichenko: I replace add to copy statements. 
To copy multiple directories in a single command (not the content), they need to have common parent dir, and copy that parent. 
Also we can replace 3 `COPY `statement here to `COPY . .`, but a lot of additional files should be added to .dockerignore.

.build/mainnet.json
```diff
@@ -0,0 +1,3 @@
+{
+  "AnchoredPriceView": "0xDECee5C06B8CC6c4D85dfd8518C246fd1497b2AE"
```
coburncoburn, line 2: need to add the data contract address

Dockerfile
```diff
@@ -15,8 +15,9 @@ COPY package.json /open-oracle/package.json
 RUN yarn install
 
 ENV PROVIDER PROVIDER
-ADD contracts contracts
-ADD saddle.config.js saddle.config.js
+COPY contracts contracts
+COPY tests tests
+COPY saddle.config.js saddle.config.js
```
jflatow, line 20: Still 3 layers - can't this be `COPY contracts tests saddle.config.js .`?

antonina-cherednichenko: not really, such copy statement will copy content of dirs, but not directories itself. We can do COPY . ., but then .dockerignore should be extended significantly.

jflatow: but so why not just use the dockerignore?

contracts/AnchoredPriceView.sol
```diff
@@ -0,0 +1,328 @@
...
+    AnchorPriceOracle immutable anchor;
+
+    /// @notice The highest ratio of the new median price to the anchor price that will still trigger the median price to be updated
+    uint256 upperBoundAnchorRatio;
+
+    /// @notice The lowest ratio of the new median price to the anchor price that will still trigger the median price to be updated
+    uint256 lowerBoundAnchorRatio;
+
+    /// @notice The mapping of posted by source prices per symbol
+    mapping(string => uint64) _prices;
```
jflatow, line 49: So what's the reason for `uint64` here now?

contracts/AnchoredPriceView.sol
jflatow: Still think we should rename input `usdcPrice` to `usdcPriceInEther` or something

contracts/AnchoredPriceView.sol
```diff
@@ -0,0 +1,328 @@
...
+    function getAnchorPrice(address tokenAddress, uint256 usdcPrice) public view returns (uint64) {
+
+        if ( tokenAddress == cUsdcAddress || tokenAddress == cUsdtAddress )  {
+            // hard code to 1 dollar
+            return uint64(oneDollar);
+        }
+
+        uint priceInEth = anchor.getUnderlyingPrice(tokenAddress);
+        uint additionalScale;
+        if ( tokenAddress == cWbtcAddress ){
```
jflatow, line 220: Minor nit: but the spacing is odd on a couple of these `if`s

contracts/AnchoredPriceView.sol
```diff
@@ -0,0 +1,331 @@
...
+     * base decimals is 1e6, so start by addint twelve
+     */
+    function getAdditionalScale(address cToken) public view returns (uint256) {
+        // total scale 1e30
+        if (cToken == cUsdcAddress) return 1e24;
+        if (cToken == cUsdtAddress) return 1e24;
+        // total scale 1e28
+        if (cToken == cWbtcAddress) return 1e22;
+        // total scale 1e18
+        if (cToken == cEthAddress) return 1e12;
```
antonina-cherednichenko, line 271: what about scale for other tokens?

coburncoburn: they aren't being posted by coinbase right now, but I think you are right maybe we can just return 1e12 for the rest of them ( they all have 18 decimals ) and this will automatically work if coinbase adds them. I was kind of in a "we'll have to change the view when they add assets" mindset, but that may not be true

contracts/AnchoredPriceView.sol
```diff
@@ -0,0 +1,331 @@
...
+     * @notice Returns price denominated in USD, with 6 decimals
+     * @dev If price was posted by source, return it. Otherwise, return anchor price converted through source ETH price.
+     */
+    function prices(string calldata symbol) external view returns (uint256) {
+        uint256 price = _prices[symbol];
+
+        if (price != 0) {
+            return price;
+        } else {
+            uint256 usdPerEth = _prices["ETH"];
```
antonina-cherednichenko, line 201: what is our source failed to post ETH price, should we have a check that price of ETH is not 0?

coburncoburn: if the source hasn't posted eth at least once, we should not use this view at all and returning 0 for every price is a reasonable result. 

A require statement and reverting may be better though?

Dockerfile
hayesgm: FWIW, we probably should make sure that dockerignore is up to date, since it sends the "build context" and adds a lot of time to building docker images if it includes unnecessary resources.

contracts/AnchoredPriceView.sol
```diff
@@ -0,0 +1,331 @@
...
+            uint256 ethPerToken = anchor.getUnderlyingPrice(getCTokenAddress(symbol));
+
+            // ethPerToken has 18 decimals since the usdt, usdc, wbtc tokens hit
```
hayesgm, line 204: I'm a bit confused by this comment

contracts/AnchoredPriceView.sol
hayesgm: Yeah, this sounds look a good place to `require`

contracts/AnchoredPriceView.sol
```diff
@@ -0,0 +1,331 @@
...
+     */
+    function getAnchorPrice(address tokenAddress, uint256 usdcPrice) public view returns (uint256) {
+        if ( tokenAddress == cUsdcAddress || tokenAddress == cUsdtAddress )  {
+            // hard code to 1 dollar
+            return oneDollar;
+        }
+
+        uint priceInEth = anchor.getUnderlyingPrice(tokenAddress);
+        uint additionalScale;
+        if ( tokenAddress == cWbtcAddress ){
```
hayesgm, line 223: Reference point for conversation below-- why doesn't additional scale come from the function below?

contracts/AnchoredPriceView.sol
```diff
@@ -0,0 +1,331 @@
...
+            return mul(ethPerToken, additionalScale);
+        }
+    }
+
+    /**
+     * comptroller expects price to have 18 decimals,
+     * additionally upscaled by 1e18 - underlyingdecimals
+     * base decimals is 1e6, so start by addint twelve
+     */
+    function getAdditionalScale(address cToken) public view returns (uint256) {
```
hayesgm, line 264: This is good-- what if we had one function called `getTokenMetadata()` and it returned a tuple or struct that the caller could parse for the relevant data? That would probably remove the repetitiveness of these functions.

package.json
hayesgm: `>=`?

contracts/AnchoredPriceView.sol
```diff
@@ -0,0 +1,331 @@
...
+        return mul(priceInEth, additionalScale) / usdcPrice;
+    }
+
+    /**
+     * @notice Implements the method of the PriceOracle interface of Compound v2 and returns returns the Eth price for an asset.
+     * @dev converts from 1e6 decimals of Open Oracle to 1e(36 - underlyingDecimals) of PriceOracleProxy
+     * @param cToken The cToken address for price retrieval
+     * @return The price for the given cToken address
+     */
+    function getUnderlyingPrice(address cToken) external view returns (uint256) {
```
hayesgm, line 241: I'm debating a bit if it's a good idea to return these in Ether versus simply returning them in USD, as we originally planned. This could probably drop some complexity (that we're going to add with other things like a staleness check)

contracts/AnchoredPriceView.sol
coburncoburn: yeah this is a slightly different additional scale, 
this one is scaling up so that dividing by USDC price ( 1e30 ) does not truncate to 0.

Yes naming is confusing and should be changed. I'm also looking at running the anchor through ETH instead so that we are only ever converting in one direction

contracts/AnchoredPriceView.sol
coburncoburn: yeah, I say we keep ruminating on which numerator to return. I'm pretty comfortable with either, have code to do either, and glad to stay nimble about which we pick in the end

PR #58, opened 2020-04-27
Toni/add anchor view fallback prices

antonina-cherednichenko:
This PR contains 2 new logic parts 
1) Implementation of this ticket https://linear.app/issue/COMP-992/have-fallback-prices-return-from-the-prices[symbol]-public-function-on after discussion with @coburncoburn 

2) Splitting `AnchoredPriceView` into 2 contracts - contract  for posting prices, anchoring and contract specific to compound protocol part with `isPriceOracle()` and `getUnderlyingPrice`. It makes AnchoredView more logically clean and readable - IMHO :)


contracts/CompoundProtocolPriceView.sol
```diff
@@ -0,0 +1,91 @@
...
+    function getUnderlyingPrice(address cToken) external view override returns (uint256) {
+        uint256 priceSixDecimals;
+
+        if(cToken == cUsdcAddress || cToken == cUsdtAddress) {
+            priceSixDecimals = oneDollar;
+        } else if (cToken == cWbtcAddress || cToken == cEthAddress) {
+            priceSixDecimals = _prices[getOracleKey(cToken)];
+        } else {
+            uint256 usdPerEth = _prices["ETH"];
+            uint256 ethPerToken = anchor.getUnderlyingPrice(cToken);
```
coburncoburn, line 51: the fact that this contract still has only this one reference to the anchor is a good indicator that writing anchor prices to storage like in the other PR is the correct abstraction. Good thinking. 

Too bad though, I don't think there is a way to pull *all* the cToken references through into this ocntract b/c of needing to know the cusdc address

antonina-cherednichenko: Yes, having tokens addresses shared between these 2 contracts is messing up this split. I kind of like splitting anchor that happens to be protocol oracle proxy from usage by protocol proxy, but it definitely needs more thought process.

PR #59, opened 2020-04-28
Cob/invalidation

coburncoburn:


contracts/AnchoredPriceView.sol
```diff
@@ -288,6 +295,15 @@ contract AnchoredPriceView {
         revert("Unknown token address");
     }
 
+    function invalidate(bytes memory message, bytes memory signature) public {
+        (string memory decoded_message, address target) = abi.decode(message, (string, address));
+        require(keccak256(abi.encodePacked(decoded_message)) == keccak256(abi.encodePacked("rotate")), "invalid message must be 'rotate'");
+        require(priceData.source(message, signature) == source, "invalidation message must come from the reporter");
+        target; //shh
```
antonina-cherednichenko, line 302: curious what this line does?

coburncoburn:
```
➜  open-oracle git:(cob/invalidation) yarn compile
yarn run v1.22.4
warning ../../package.json: No license field
$ npx saddle compile
Compiling contracts contracts/*.sol tests/contracts/*.sol with solc to /Users/coburn/Compound/open-oracle/.build/contracts.json...
Warning: Unused local variable.
   --> contracts/AnchoredPriceView.sol:299:41:
    |
299 |         (string memory decoded_message, address target) = abi.decode(message, (string, address));
    |                                         ^^^^^^^^^^^^^^


```
just to get rid of this unused local variable complaint : /

contracts/AnchoredPriceView.sol
```diff
@@ -221,6 +225,9 @@ contract AnchoredPriceView {
      * @return The price for the given cToken address
      */
     function getUnderlyingPrice(address cToken) external view returns (uint256) {
+        if (breaker == true) {
+            return anchor.getUnderlyingPrice(cToken);
```
antonina-cherednichenko, line 229: so, if breaker is on, price in ETH is returned, and if it's off - USD?

coburncoburn: this is true, here, but in anticipation of going for only ETH, right?

contracts/AnchoredPriceView.sol
antonina-cherednichenko: wow 🤩

coburncoburn:
@antonina-cherednichenko I don't think this iteration makes sense to be able to turn breaker back off, since the source key writing an invalidation message means , in worst case, they have lost the signing key and a newly configured view would be the way to go in that situation. 

Having said that, the key rotation story is pretty much in the minimal form at this point ( want to rotate keys? get a new view ), so something to keep thinking about for future iterations.

contracts/AnchoredPriceView.sol
jflatow: I think on this one you could also do `(string memory decoded_message, ) = ...`

coburncoburn: perfect, I tried (string memory, address ) and that failed so I gave up. Your take works though

PR #61, opened 2020-04-30
Cob/always eth

coburncoburn:


Dockerfile
```diff
@@ -15,8 +15,9 @@ COPY package.json /open-oracle/package.json
 RUN yarn install
 
 ENV PROVIDER PROVIDER
-ADD contracts contracts
-ADD saddle.config.js saddle.config.js
+COPY contracts contracts
+COPY tests tests
+COPY saddle.config.js saddle.config.js
```
jflatow, line 20: Still 3 layers, can't this just be `COPY contracts tests saddle.config.js .`?

PR #62, opened 2020-04-30
have get underying return values in eth, and have prices return value…

coburncoburn:
…s in usd, pivoting on source ETH price rather than anchor

### DIFF
--- a/contracts/AnchoredPriceView.sol
+++ b/contracts/AnchoredPriceView.sol
@@ -15,13 +15,13 @@ interface AnchorPriceOracle {
  */
 contract AnchoredPriceView {
     /// @notice standard amount for the Dollar
-    uint64 constant oneDollar = 1e6;
+    uint256 constant oneDollar = 1e6;
 
     /// @notice The event emitted when the median price is updated
-    event PriceUpdated(string symbol, uint64 price);
+    event PriceUpdated(string symbol, uint256 price);
 
-    /// @notice The event emitted when new prices are posted but the median price is not updated due to the anchor
-    event PriceGuarded(string symbol, uint64 median, uint64 anchor);
+    /// @notice The event emitted when new prices are posted but the stored price is not updated due to the anchor
+    event PriceGuarded(string symbol, uint256 source, uint256 anchor);
 
     /// @notice The CToken contracts addresses
     struct CTokens {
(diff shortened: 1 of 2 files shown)


contracts/AnchoredPriceView.sol
```diff
@@ -46,7 +46,7 @@ contract AnchoredPriceView {
     uint256 lowerBoundAnchorRatio;
 
     /// @notice The mapping of posted by source prices per symbol
-    mapping(string => uint64) _prices;
+    mapping(string => uint64) public _prices;
```
jflatow, line 49: What's the benefit of `uint64` since these don't really save gas?

tests/AnchoredPriceViewTest.js
```diff
@@ -326,46 +326,47 @@ describe('AnchoredPriceView', () => {
...
+        ],
+        ['ETH', 'BTC'],
+        source
+      );
     });
 
 
     ["USDC", "USDT"].forEach((openOracleKey) => {
-      it(`returns 1 with 18 + 12 decimals for ${openOracleKey}`, async () => {
+      it(`returns 1 converted to ETH throush source ETH with 18 + 12 decimals for ${openOracleKey}`, async () => {
```
jflatow, line 352: through?

contracts/AnchoredPriceView.sol
coburncoburn: I am good to go with 256

contracts/AnchoredPriceView.sol
```diff
@@ -185,17 +187,27 @@ contract AnchoredPriceView {
...
         } else {
-            uint256 usdcPrice = anchor.getUnderlyingPrice(cUsdcAddress);
-            return getAnchorPrice(getCTokenAddress(symbol), usdcPrice);
+            uint256 usdPerEth = _prices["ETH"];
+            uint256 ethPerToken = anchor.getUnderlyingPrice(getCTokenAddress(symbol));
+
+            // ethPerToken has 18 decimals since the usdt, usdc, wbtc tokens hit
+            // usdPerEth has 6 decimals
+            // scaling by 1e18 and dividing leaves 1e6, as desired
+            return uint64(mul(usdPerEth, 1e18) / ethPerToken);
```
coburncoburn, line 206: probably won't matter in the real world, but I thought it was tighter for the "source" eth price to control the values here rather than anchor "usdc" price

PR #66, opened 2020-05-07
 Cob/deadman cut anchor

coburncoburn:
note, sai tests are failing right now b/c they have a hardcoded value for sai price that will need to be set ( on May 12, allegedly ), so rather keep those tests failing 

1. Split View into two contracts, one for all the hairy config, other for the logic. 
  - Hairy config could still be untangled / verified in constructor.
2. Add cutAnchor function that checks if last v1 oracle update for usdc was over 5760 blocks in the past ( ~ 24 hours , maybe we should bump up in case of fast blocks ). If it is, don't block updates from CB.
3. Put the current proxy logic into the view, since we now must depend on v1 oracle direclty to get the time periods, might as well pull that function in

### DIFF
--- a/contracts/AnchoredView/AnchoredView.sol
+++ b/contracts/AnchoredView/AnchoredView.sol
@@ -4,99 +4,93 @@ pragma experimental ABIEncoderV2;
 import "./SymbolConfiguration.sol";
 import "../OpenOraclePriceData.sol";
 
-interface CErc20 {
-    function underlying() external view returns (address);
-}
-
 interface AnchorOracle {
     function numBlocksPerPeriod() external view returns (uint); // approximately 1 hour: 60 seconds/minute * 60 minutes/hour * 1 block/15 seconds
 
     function assetPrices(address asset) external view returns (uint);
 
+    struct Anchor {
+        // floor(block.number / numBlocksPerPeriod) + 1
+        uint period;
 
-    /* struct Anchor { */
-    /*     // floor(block.number / numBlocksPerPeriod) + 1 */
-    /*     uint period; */
-
-    /*     // Price in ETH, scaled by 10**18 */
-    /*     uint priceMantissa; */
-    /* } */
-    function anchors(address asset) external view returns (uint, uint);
+        // Price in ETH, scaled by 10**18
+        uint priceMantissa;
+    }
+    function anchors(address asset) external view returns (Anchor memory);
 }
 
 
 /**
  * @notice Price feed conforming to Price Oracle Proxy interface.
  * @dev Use a single open oracle reporter and anchored to and falling back to the Compound v2 oracle system.
+ * @dev The reporter must report at a minimum the USD/ETH price, so that anchor ETH/TOKEN prices can be converted to USD/TOKEN
  * @author Compound Labs, Inc.
  */
 contract AnchoredView is SymbolConfiguration {
-    /// @notice The mapping of posted by source prices per symbol
-    mapping(string => uint256) public _prices;
+    /// @notice The mapping of anchored reporter prices by symbol
+    mapping(string => uint) public _prices;
 
-    /// @notice circuit breaker for using anchor price oracle directly
-    bool public breaker;
+    /// @notice Circuit breaker for using anchor price oracle directly, ignoring reporter
+    bool public reporterBreaker;
 
-    /// @notice circuit breaker for using source price without anchor
-    bool public anchored = true;
+    /// @notice Circuit breaker for using reporter price without anchor
+    bool public anchorBreaker;
 
-    /// @notice the Open Oracle Reporter price source
-    address public immutable source;
+    /// @notice the Open Oracle Reporter price reporter
+    address public immutable reporter;
 
-    /// @notice the anchor oracle ( Compouni Oracle V1 )
+    /// @notice The anchor oracle ( Compound Oracle V1 )
     AnchorOracle public immutable anchor;
 
-    /// @notice the Open Oracle Price Data contract
+    /// @notice The Open Oracle Price Data contract
     OpenOraclePriceData public immutable priceData;
 
     /// @notice The highest ratio of the new median price to the anchor price that will still trigger the median price to be updated
-    uint256 immutable upperBoundAnchorRatio;
+    uint immutable upperBoundAnchorRatio;
 
     /// @notice The lowest ratio of the new median price to the anchor price that will still trigger the median price to be updated
-    uint256 immutable lowerBoundAnchorRatio;
+    uint immutable lowerBoundAnchorRatio;
 
-    /// @notice standard amount for the Dollar
-    uint256 constant oneDollar = 1e6;
+    /// @notice Average blocks per day, for checking anchor staleness
+    /// @dev 1 day / 15
+    uint constant blocksInADay = 5760;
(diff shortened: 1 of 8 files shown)


contracts/AnchoredView/AnchoredView.sol
```diff
@@ -0,0 +1,278 @@
...
+     * @param anchorToleranceMantissa_ The tolerance allowed between the anchor and median. A tolerance of 10e16 means a new median that is 10% off from the anchor will still be saved
+     * @param tokens_ The CTokens struct that contains addresses for CToken contracts
+     */
+    constructor(OpenOraclePriceData data_,
+                address source_,
+                address anchor_,
+                uint anchorToleranceMantissa_,
+                CTokens memory tokens_) SymbolConfiguration(tokens_) public {
+        source = source_;
+        anchor = AnchorOracle(anchor_);
```
coburncoburn, line 86: can wrap this in arg list instead of here

contracts/AnchoredView/AnchoredView.sol
```diff
@@ -0,0 +1,278 @@
...
+    function anchors(address asset) external view returns (uint, uint);
+}
+
+
+/**
+ * @notice Price feed conforming to Price Oracle Proxy interface.
+ * @dev Use a single open oracle reporter and anchored to and falling back to the Compound v2 oracle system.
+ * @author Compound Labs, Inc.
+ */
+contract AnchoredView is SymbolConfiguration {
```
antonina-cherednichenko, line 33: should AnchoredPriceView.sol be deleted?

coburncoburn: yeah, may steal that contract name too. I was depending on the old tests passing to see if I broke things, but if this version looks good, will delete the old one

contracts/AnchoredView/AnchoredView.sol
```diff
@@ -0,0 +1,278 @@
...
+pragma experimental ABIEncoderV2;
+
+import "./SymbolConfiguration.sol";
+import "../OpenOraclePriceData.sol";
+
+interface CErc20 {
+    function underlying() external view returns (address);
+}
+
+interface AnchorOracle {
```
antonina-cherednichenko, line 11: is there an example of an anchor that will be used ?

coburncoburn: great question. 
 https://etherscan.io/address/0x02557a5e05defeffd4cae6d83ea3d173b272c904#code
https://github.com/compound-finance/compound-oracle/blob/master/contracts/PriceOracle.sol

contracts/AnchoredView/AnchoredView.sol
```diff
@@ -0,0 +1,278 @@
...
+
+        breaker = true;
+        emit SourceInvalidated(source);
+    }
+
+    /// @notice invalidate the anchor, and fall back to using source without anchor
+
+    /// @dev determine if anchor is stale by checking when usdc was last updated
+    // @dev all anchor prices are converted through usdc price, so if it is stale they are all stale
+    function cutAnchor() public {
```
antonina-cherednichenko, line 253: how often this method will be called?

coburncoburn: it will only do anything once, in the event the current oracle goes down for 24 hours ( or maybe there is a problem here where it would also be possible if ETH / USDC was completely stable for 24 hours )

coburncoburn: we could also inline this code for "ignoreAnchor" and call it everytime the price is outside of anchor rather than cutting entirely, so that if anchor comes back online, it is used again.

contracts/AnchoredView/AnchoredView.sol
```diff
@@ -0,0 +1,278 @@
...
+     * @notice Returns price denominated in USD, with 6 decimals
+     * @dev If price was posted by source, return it. Otherwise, return anchor price converted through source ETH price.
+     */
+    function prices(string calldata symbol) external view returns (uint256) {
+        uint256 price = _prices[symbol];
+
+        if (price != 0) {
+            return price;
+        } else {
+            uint256 usdPerEth = _prices["ETH"];
```
hayesgm, line 152: What if this is zero?

contracts/AnchoredView/AnchoredView.sol
```diff
@@ -0,0 +1,278 @@
...
+    /// @dev determine if anchor is stale by checking when usdc was last updated
+    // @dev all anchor prices are converted through usdc price, so if it is stale they are all stale
+    function cutAnchor() public {
+        (uint latestUsdcAnchorPeriod,) = anchor.anchors(usdcOracleKey);
+
+        uint usdcAnchorBlockNumber = mul(latestUsdcAnchorPeriod, anchor.numBlocksPerPeriod());
+        uint blocksSinceUpdate = block.number - usdcAnchorBlockNumber;
+
+        // one day in 15 second blocks without an update
+        if (blocksSinceUpdate > 5760) {
```
hayesgm, line 260: Can we extract this to a constant?

contracts/AnchoredView/SymbolConfiguration.sol
```diff
@@ -0,0 +1,119 @@
...
+        cSaiAddress = tokens_.cSaiAddress;
+        cUsdtAddress = tokens_.cUsdtAddress;
+    }
+
+    /**
+     * comptroller expects price to have 18 decimals,
+     * additionally upscaled by 1e18 - underlyingdecimals
+     * base decimals is 1e6, so start by addint twelve
+     */
+    function getAdditionalScale(address cToken) public view returns (uint256) {
```
hayesgm, line 72: These are fine, though I still think it might be even easier to simply return a tuple and then we don't need to repeat this logic multiple times. Also, should this be able to get rid of the other additional scale and one-offs used in the anchor contract above? Like, it'd be cool if there were the (generally) only metadata function(s) and thus we could get rid of statements like: `if (cTokenAddress == cUsdcAddress || cTokenAddress == cUsdtAddress) {`

contracts/AnchoredView/SymbolConfiguration.sol
```diff
@@ -0,0 +1,119 @@
+pragma solidity ^0.6.6;
+pragma experimental ABIEncoderV2;
+
+// verbose configuration of anchored view, transforming
+// address to symbols and such
+contract SymbolConfiguration {
```
hayesgm, line 6: I do want to be weary about hiding complexity here. Is the advantage we're getting here simply that the code lives in a separate file?

contracts/AnchoredView/AnchoredView.sol
antonina-cherednichenko, line 260: maybe worth moving it to constant or even setting in constructor?

contracts/AnchoredView/SymbolConfiguration.sol
coburncoburn: yeah, it was just a distraction for navigating the file. Happy to inline it later when not so much development happening

contracts/AnchoredView/AnchoredView.sol
coburncoburn: then we return 0, could revert instead.

PR #67, opened 2020-05-12
Cob/ctoken config struct

coburncoburn:
big refactor on the config, 
back to returning usd instead of eth in get underlying
I would consider this branch mostly final for this iteration of it all, pending any reviews

### DIFF
--- a/contracts/AnchoredView/AnchoredView.sol
+++ b/contracts/AnchoredView/AnchoredView.sol
@@ -164,6 +164,7 @@ contract AnchoredView is SymbolConfiguration {
      * @dev Anchor usdc price has 30 decimals, and anchor general price has 18 decimals, so multiplying 1e18 by 1e18 and dividing by 1e30 yields 1e6
      */
     function getAnchorInUsd(address cToken, uint ethPerUsdc) public view returns (uint) {
+        if (cToken == cUsdcAddress || cToken == cUsdtAddress) return 1e6;
         CTokenMetadata memory tokenConfig = getCTokenConfig(cToken);
         uint ethPerToken = readAnchor(tokenConfig);
 
@@ -177,7 +178,7 @@ contract AnchoredView is SymbolConfiguration {
      * @param cToken The cToken address for price retrieval
      * @return The price for the given cToken address
      */
-    function getUnderlyingPrice(address cToken) public view returns (uint) {
+    function getUnderlyingPrice(address cToken) public returns (uint) {
         CTokenMetadata memory tokenConfig = getCTokenConfig(cToken);
         if (breaker == true) {
             return readAnchor(tokenConfig);
@@ -195,7 +196,9 @@ contract AnchoredView is SymbolConfiguration {
 
             // factoring out extra 6 decimals from source eth price brings us back to decimals given by anchor
             uint ethPerToken = readAnchor(tokenConfig);
-            usdPerToken = mul(usdPerEth, ethPerToken) / 1e6;
+            emit PriceUpdated("eth per token", ethPerToken);
+            emit PriceUpdated("usd per eth", usdPerEth);
+            return mul(usdPerEth, ethPerToken) / 1e6;
         }
     }
 
--- a/contracts/AnchoredView/AnchoredView.sol
+++ b/contracts/AnchoredView/AnchoredView.sol
@@ -27,22 +27,22 @@ interface AnchorOracle {
  * @author Compound Labs, Inc.
  */
 contract AnchoredView is SymbolConfiguration {
-    /// @notice The mapping of posted by reporter prices per symbol
+    /// @notice The mapping of anchored reporter prices by symbol
     mapping(string => uint) public _prices;
 
-    /// @notice circuit breaker for using anchor price oracle directly
+    /// @notice Circuit breaker for using anchor price oracle directly
     bool public breaker;
 
-    /// @notice circuit breaker for using reporter price without anchor
+    /// @notice Circuit breaker for using reporter price without anchor
     bool public anchored = true;
 
     /// @notice the Open Oracle Reporter price reporter
     address public immutable reporter;
 
-    /// @notice the anchor oracle ( Compouni Oracle V1 )
+    /// @notice The anchor oracle ( Compound Oracle V1 )
     AnchorOracle public immutable anchor;
 
-    /// @notice the Open Oracle Price Data contract
+    /// @notice The Open Oracle Price Data contract
     OpenOraclePriceData public immutable priceData;
 
     /// @notice The highest ratio of the new median price to the anchor price that will still trigger the median price to be updated
@@ -51,7 +51,7 @@ contract AnchoredView is SymbolConfiguration {
     /// @notice The lowest ratio of the new median price to the anchor price that will still trigger the median price to be updated
     uint immutable lowerBoundAnchorRatio;
 
-    /// @notice average blocks per day, for checking anchor staleness
+    /// @notice Average blocks per day, for checking anchor staleness
     /// @dev 1 day / 15
     uint constant blocksInADay = 5760;
 
@@ -230,7 +230,7 @@ contract AnchoredView is SymbolConfiguration {
     /// @notice invalidate the anchor, and fall back to using reporter without anchor
 
     /// @dev determine if anchor is stale by checking when usdc was last updated
-    // @dev all anchor prices are converted through usdc price, so if it is stale they are all stale
+    /// @dev all anchor prices are converted through usdc price, so if it is stale they are all stale
     function cutAnchor() external {
         (uint latestUsdcAnchorPeriod,) = anchor.anchors(cUsdcAnchorKey);
 
--- a/contracts/AnchoredView/AnchoredView.sol
+++ b/contracts/AnchoredView/AnchoredView.sol
@@ -4,10 +4,6 @@ pragma experimental ABIEncoderV2;
 import "./SymbolConfiguration.sol";
 import "../OpenOraclePriceData.sol";
 
-interface CErc20 {
-    function underlying() external view returns (address);
-}
-
 interface AnchorOracle {
     function numBlocksPerPeriod() external view returns (uint); // approximately 1 hour: 60 seconds/minute * 60 minutes/hour * 1 block/15 seconds
 
@@ -32,7 +28,7 @@ interface AnchorOracle {
  */
 contract AnchoredView is SymbolConfiguration {
     /// @notice The mapping of posted by source prices per symbol
-    mapping(string => uint256) public _prices;
+    mapping(string => uint) public _prices;
 
     /// @notice circuit breaker for using anchor price oracle directly
     bool public breaker;
@@ -50,23 +46,23 @@ contract AnchoredView is SymbolConfiguration {
     OpenOraclePriceData public immutable priceData;
 
     /// @notice The highest ratio of the new median price to the anchor price that will still trigger the median price to be updated
-    uint256 immutable upperBoundAnchorRatio;
+    uint immutable upperBoundAnchorRatio;
 
     /// @notice The lowest ratio of the new median price to the anchor price that will still trigger the median price to be updated
-    uint256 immutable lowerBoundAnchorRatio;
+    uint immutable lowerBoundAnchorRatio;
 
     /// @notice standard amount for the Dollar
-    uint256 constant oneDollar = 1e6;
+    uint constant oneDollar = 1e6;
 
     /// @notice average blocks per day, for checking anchor staleness
     /// @dev 1 day / 15
-    uint256 constant blocksInADay = 5760;
+    uint constant blocksInADay = 5760;
 
     /// @notice The event emitted when the median price is updated
-    event PriceUpdated(string symbol, uint256 price);
+    event PriceUpdated(string symbol, uint price);
 
     /// @notice The event emitted when new prices are posted but the stored price is not updated due to the anchor
-    event PriceGuarded(string symbol, uint256 source, uint256 anchor);
+    event PriceGuarded(string symbol, uint source, uint anchor);
 
     /// @notice The event emitted when source invalidates itself
     event SourceInvalidated(address source);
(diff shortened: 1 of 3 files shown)


contracts/AnchoredView/AnchoredView.sol
```diff
@@ -143,45 +143,33 @@ contract AnchoredView is SymbolConfiguration {
...
-        uint additionalScale;
-        if ( tokenAddress == cWbtcAddress ){
-            // wbtc proxy price is scaled 1e(36 - 8) = 1e28, so we need 8 more to get to 36
-            additionalScale = 1e8;
-        } else {
-            // all other tokens are scaled 1e18, so we need 18 more to get to 36
-            additionalScale = 1e18;
-        }
+    function getAnchorInUsd(address cToken, uint ethPerUsdc) public view returns (uint) {
+        if (cToken == cUsdcAddress || cToken == cUsdtAddress) return 1e6;
```
jflatow, line 167: Isn't this just because the tokens have 6 decimals, I think the comment might make it more confusing?

contracts/AnchoredView/SymbolConfiguration.sol
```diff
@@ -1,17 +1,28 @@
...
-    address public constant daiOracleKey = address(2);
+    /// Address of the oracle key (underlying) for cTokens non special keyed tokens
+    address public immutable cRepAnchorKey;
+    address public immutable cWbtcAnchorKey;
+    address public immutable cBatAnchorKey;
+    address public immutable cZrxAnchorKey;
 
-    /// @notice Frozen SAI price in ETH
-    uint public saiPrice = 1e18;
+    /// Frozen rices for SAI and eth, so no oracle key
```
jflatow, line 23: prices?

contracts/AnchoredView/SymbolConfiguration.sol
```diff
@@ -1,17 +1,28 @@
...
+    uint public saiAnchorPrice = 5285551943761727;
+    uint public ethAnchorPrice = 1e18;
```
jflatow, line 25: should these be constant?

contracts/AnchoredView/AnchoredView.sol
coburncoburn: this is b/c cusdc and cusdt never change, In reality this won't ever get called internally with cusdc and cusdt. We could actually delete this line, actually it should always return 1e6 anwyays after running the math for real. 

Also, it could be internal ...

contracts/AnchoredView/SymbolConfiguration.sol
coburncoburn: yes absolutley

contracts/AnchoredView/AnchoredView.sol
coburncoburn: or if you mean the comment below, we are converting through the current price oracle eth usd price, and it has 30 decimals. Whatever the numerator is, we need to make sure it has 36 decimals before dividing, ending with the 6 we want

contracts/AnchoredView/AnchoredView.sol
```diff
@@ -110,29 +110,29 @@ contract AnchoredView is SymbolConfiguration {
...
+            if (tokenAddress == cUsdcAddress || tokenAddress == cUsdtAddress || anchorPrice == 0)  {
                 emit PriceGuarded(symbol, sourcePrice, anchorPrice);
             } else {
-                uint256 anchorRatioMantissa = mul(sourcePrice, 100e16) / anchorPrice;
-                // Only update the view's price if the source price is within a bound
-                if (anchorRatioMantissa <= upperBoundAnchorRatio && anchorRatioMantissa >= lowerBoundAnchorRatio) {
+                uint anchorRatio = mul(sourcePrice, 100e16) / anchorPrice;
+                bool withinAnchor = anchorRatio <= upperBoundAnchorRatio && anchorRatio >= lowerBoundAnchorRatio;
+
+                if (withinAnchor || !anchored) {
```
antonina-cherednichenko, line 130: Maybe `!anchored` check and price update can be moved even above `getAnchorInUsd(tokenAddress, usdcPrice);` ? 
Otherwise all mantissa checks are done for nothing if it's !anchored

contracts/AnchoredView/AnchoredView.sol
antonina-cherednichenko: maybe oneDollar constant instead of 1e6 for clarity?

contracts/AnchoredView/AnchoredView.sol
```diff
@@ -190,50 +178,50 @@ contract AnchoredView is SymbolConfiguration {
      * @param cToken The cToken address for price retrieval
      * @return The price for the given cToken address
      */
-    function getUnderlyingPrice(address cToken) external view returns (uint256) {
+    function getUnderlyingPrice(address cToken) public returns (uint) {
+        CTokenMetadata memory tokenConfig = getCTokenConfig(cToken);
         if (breaker == true) {
-            return readAnchor(cToken);
+            return readAnchor(tokenConfig);
```
antonina-cherednichenko, line 184: should it be have the same adjustment of precision as in the last case : `mul(usdPerEth, ethPerToken) / 1e6;` ?

contracts/AnchoredView/AnchoredView.sol
```diff
@@ -251,26 +239,26 @@ contract AnchoredView is SymbolConfiguration {
     /// @dev determine if anchor is stale by checking when usdc was last updated
     // @dev all anchor prices are converted through usdc price, so if it is stale they are all stale
     function cutAnchor() public {
```
antonina-cherednichenko, line 241: public vs external here

contracts/AnchoredView/AnchoredView.sol
```diff
@@ -190,50 +178,50 @@ contract AnchoredView is SymbolConfiguration {
...
-            return readAnchor(cToken);
+            return readAnchor(tokenConfig);
         }
 
-        uint256 usdPerToken = _prices[getOracleKey(cToken)];
+        uint usdPerToken = _prices[tokenConfig.openOracleKey];
 
-        if ( usdPerToken == 0 ) {
-            return readAnchor(cToken);
+        if ( usdPerToken != 0 ) {
```
antonina-cherednichenko, line 189: probably no spaces here, in if clause?

contracts/AnchoredView/AnchoredView.sol
```diff
@@ -190,50 +178,50 @@ contract AnchoredView is SymbolConfiguration {
      * @param cToken The cToken address for price retrieval
      * @return The price for the given cToken address
      */
-    function getUnderlyingPrice(address cToken) external view returns (uint256) {
+    function getUnderlyingPrice(address cToken) public returns (uint) {
```
antonina-cherednichenko, line 181: what happens, if anchor is stale here?

contracts/AnchoredView/AnchoredView.sol
coburncoburn: up for discussion, but seems like falling back all the way is winning the day right now. Just pretend this whole thing never happened and serve prices in ether from v1 oracle

contracts/AnchoredView/AnchoredView.sol
coburncoburn: if anchor is stale here, kind of doomed. We really can't let anchor go down until Reporters cover all markets

contracts/AnchoredView/AnchoredView.sol
```diff
@@ -27,17 +27,17 @@ interface AnchorOracle {
  * @author Compound Labs, Inc.
  */
 contract AnchoredView is SymbolConfiguration {
-    /// @notice The mapping of posted by source prices per symbol
+    /// @notice The mapping of posted by reporter prices per symbol
```
jflatow, line 30: Isn't this just the mapping of (anchored, medianized, etc.) prices by symbol?

coburncoburn: tru

contracts/AnchoredView/AnchoredView.sol
```diff
@@ -27,17 +27,17 @@ interface AnchorOracle {
...
-    /// @notice circuit breaker for using source price without anchor
+    /// @notice circuit breaker for using reporter price without anchor
     bool public anchored = true;
 
-    /// @notice the Open Oracle Reporter price source
-    address public immutable source;
+    /// @notice the Open Oracle Reporter price reporter
+    address public immutable reporter;
 
     /// @notice the anchor oracle ( Compouni Oracle V1 )
```
jflatow, line 42: Nit: `Compouni`, and I think the notices should always be capitalized?

contracts/AnchoredView/AnchoredView.sol
coburncoburn: to question 1, at first !anchored was it's own block after this one since I was trying to skip that storage read if it was withinAnchor, that should probably go back to how it was.

question 2, also a good point..

jflatow:
What is this? https://github.com/compound-finance/open-oracle/blob/72914811e5de4f2e7d5705aaf32ee440c0918324/contracts/AnchoredView/AnchoredView.sol#L13-L19

coburncoburn:
@jflatow that is the anchor struct in v1 oracle. There in the comment for reference, but it's to detect anchor being stale and cuttable. Could also assign to the struct when using to make it more clear, but we only want the period.

contracts/AnchoredView/SymbolConfiguration.sol
```diff
@@ -14,15 +14,18 @@ contract SymbolConfiguration {
...
     /// Address of the oracle key (underlying) for cTokens non special keyed tokens
     address public immutable cRepAnchorKey;
     address public immutable cWbtcAnchorKey;
     address public immutable cBatAnchorKey;
     address public immutable cZrxAnchorKey;
 
-    /// Frozen rices for SAI and eth, so no oracle key
-    uint public saiAnchorPrice = 5285551943761727;
-    uint public ethAnchorPrice = 1e18;
+    /// Frozen prices for SAI and eth, so no oracle key
```
jflatow, line 26: FYI I think /// is only for natspec comments

coburncoburn: I was wondering why it was 3...

contracts/AnchoredView/SymbolConfiguration.sol
```diff
@@ -102,49 +111,123 @@ contract SymbolConfiguration {
...
-    }
+        if (cToken == cEthAddress) {
+            return CTokenMetadata({
+                        openOracleKey: "ETH",
+                        anchorOracleKey: address(0),
+                        baseUnit: 1e18,
+                        cTokenAddress: cEthAddress,
+                        priceSource: PriceSource.REPORTER,
+                        anchorSource: AnchorSource.FIXED_ETH,
+                        fixedReporterPrice: 0,
```
jflatow, line 122: I assume you are being forced to initialize these to 0, even though they aren't used?

contracts/AnchoredView/AnchoredView.sol
```diff
@@ -118,58 +115,58 @@ contract AnchoredView is SymbolConfiguration {
 
             address tokenAddress = getCTokenAddress(symbol);
 
-            uint sourcePrice = priceData.getPrice(source, symbol);
+            uint reporterPrice = priceData.getPrice(reporter, symbol);
             uint anchorPrice = getAnchorInUsd(tokenAddress, usdcPrice);
 
             if (tokenAddress == cUsdcAddress || tokenAddress == cUsdtAddress || anchorPrice == 0)  {
```
jflatow, line 121: Is this because of fixed prices - I don't think this should check for `cUsdcAddress` or `cUsdtAddress`? Its just a place that can get easily out of sync, now that the config is elsewhere, and this should still return the fixed price anyway? Plus, theoretically if a reporter did want to post those prices, I think it might not be totally unreasonable. Could be in anticipation of a new view coming, or another view, or something

contracts/AnchoredView/SymbolConfiguration.sol
coburncoburn: yes, no choice :(

contracts/AnchoredView/AnchoredView.sol
coburncoburn: it is b/c of fixed prices, good point. There is another issue here that you can only post prices for configured ssets to this view. Maybe there should be an "unconfigured" configuration that writes to data contract, but skips view storage

contracts/AnchoredView/AnchoredView.sol
```diff
@@ -27,17 +27,17 @@ interface AnchorOracle {
  * @author Compound Labs, Inc.
  */
 contract AnchoredView is SymbolConfiguration {
-    /// @notice The mapping of posted by source prices per symbol
+    /// @notice The mapping of posted by reporter prices per symbol
     mapping(string => uint) public _prices;
 
     /// @notice circuit breaker for using anchor price oracle directly
     bool public breaker;
```
jflatow, line 34: Bit confusing to me that these are both breakers but one is called `breaker`

contracts/AnchoredView/AnchoredView.sol
```diff
@@ -118,58 +115,58 @@ contract AnchoredView is SymbolConfiguration {
...
     function prices(string calldata symbol) external view returns (uint) {
-        uint price = _prices[symbol];
+        CTokenMetadata memory tokenConfig = getCTokenConfig(symbol);
 
-        if (price != 0) {
-            return price;
-        } else {
+        if (tokenConfig.priceSource == PriceSource.REPORTER) return _prices[symbol];
+        if (tokenConfig.priceSource == PriceSource.FIXED_USD) return tokenConfig.fixedReporterPrice;
+        if (tokenConfig.priceSource == PriceSource.ANCHOR) {
```
jflatow, line 148: I feel like its taken for granted that anchor *must* provide prices in ETH, maybe this should be stated somewhere up front that this is the assumption?

contracts/AnchoredView/AnchoredView.sol
jflatow: Hmm separate function might be overkill, but where does it err out if its 'unconfigured'? It reverts getting the config which is inside `getAnchorInUsd`?

jflatow:
> @jflatow that is the anchor struct in v1 oracle. There in the comment for reference, but it's to detect anchor being stale and cuttable. Could also assign to the struct when using to make it more clear, but we only want the period.

Would it be so bad to uncomment and use the struct as the return value?

contracts/AnchoredView/AnchoredView.sol
coburncoburn: exactly

contracts/AnchoredView/AnchoredView.sol
coburncoburn: actually , point 1, this boolean short circuit will skip the storage read on !anchored in most cases, so I think the logic is best there.

refactoring to move the 0 check into the within anchor check so we still get the benefit of the cut anchor

contracts/AnchoredView/AnchoredView.sol
coburncoburn: added a comment on the contract natspec

contracts/AnchoredView/AnchoredView.sol
coburncoburn: renamed

contracts/AnchoredView/AnchoredView.sol
coburncoburn: tried it out with the old null-object pattern, I think the "unconfigured" metadata is pretty nice actually

PR #68, opened 2020-05-18
Anchored view

coburncoburn:


contracts/AnchoredView/AnchoredView.sol
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    function readAnchor(CTokenMetadata memory tokenConfig) internal view returns (uint) {
+        if (tokenConfig.anchorSource == AnchorSource.FIXED_ETH) return tokenConfig.fixedAnchorPrice;
+
+        return anchor.assetPrices(tokenConfig.anchorOracleKey);
+    }
+
+    /// @notice invalidate the reporter, and fall back to using anchor directly in all cases
+    function invalidate(bytes memory message, bytes memory signature) public {
```
antonina-cherednichenko, line 218: public vs external here?

antonina-cherednichenko:
really nice and clean version, great 👍 work

.build/mainnet.json
```diff
@@ -0,0 +1,4 @@
+{
+  "AnchoredPriceView": "0xDECee5C06B8CC6c4D85dfd8518C246fd1497b2AE",
```
antonina-cherednichenko, line 2: is this still up-to-date? should AnchoredView  `0x4345d279E3aeA247Ad7865cC95BeE3088eaBc433` be added?

coburncoburn: yes, forgot topush, thanks

.build/mainnet.json
```diff
@@ -0,0 +1,5 @@
...
```
antonina-cherednichenko, line 2: Is AnchoredPriceView still needed here?

jflatow: Only if its being used by the API

coburncoburn: it is not

contracts/AnchoredView/AnchoredView.sol
```diff
@@ -0,0 +1,254 @@
...
+                address reporter_,
+                AnchorOracle anchor_,
+                uint anchorToleranceMantissa_,
+                CTokens memory tokens_) SymbolConfiguration(tokens_) public {
+        reporter = reporter_;
+        anchor = anchor_;
+        priceData = data_;
+
+        require(anchorToleranceMantissa_ < 100e16, "Anchor Tolerance is too high");
+        upperBoundAnchorRatio = 100e16 + anchorToleranceMantissa_;
```
maxwolff, line 87: should we call it `upperBoundAnchorRatio` for consistency?

contracts/AnchoredView/AnchoredView.sol
```diff
@@ -0,0 +1,254 @@
...
+        }
+
+        // load usdc for using in loop to convert anchor prices to dollars
+        uint usdcPrice = readAnchor(cUsdcAddress);
+
+        // Try to update the view storage
+        for (uint i = 0; i < symbols.length; i++) {
+            CTokenMetadata memory tokenConfig = getCTokenConfig(symbols[i]);
+            // symbol is not supported in the view, but allow writing to data
+            if (tokenConfig.cTokenAddress == address(0)) continue;
```
maxwolff, line 113: is this just to explicity call out this case? doesnt impact control flow at all, right?

coburncoburn: the call will eventually revert if the token is not configured, so this is to short circuit before that happens

coburncoburn: so, yes, it does influence control flow by skipping the innards of this loop for this tokenConfig

jflatow:
I don't think we should merge this into `master` yet, I think we should prepare a complete release branch with everything we are planning to use for the live oracle

contracts/AnchoredView/AnchoredView.sol
```diff
@@ -0,0 +1,254 @@
...
+import "../OpenOraclePriceData.sol";
+
+interface AnchorOracle {
+    function numBlocksPerPeriod() external view returns (uint); // approximately 1 hour: 60 seconds/minute * 60 minutes/hour * 1 block/15 seconds
+
+    function assetPrices(address asset) external view returns (uint);
+
+    struct Anchor {
+        // floor(block.number / numBlocksPerPeriod) + 1
+        uint period;
```
maxwolff, line 14: do we only need period for checking the staleness of USDC? can we get rid of this struct just by storing a simple `lastUpdated` somewhere?

coburncoburn: this anchor struct is super specific to the version depending on the original oracle. It should not be used in the new view

jflatow:
Deprecating this PR for #70

PR #74, opened 2020-06-17
Toni/uniswap oracles

antonina-cherednichenko:
Initial version of 'simple' Uniswap Oracle for all supported pairs. It can be partially reused for Sliding window oracle.


contracts/UniswapContracts.sol
```diff
@@ -0,0 +1,281 @@
...
+    function sub(uint x, uint y) internal pure returns (uint z) {
+        require((z = x - y) <= x, 'ds-math-sub-underflow');
+    }
+
+    function mul(uint x, uint y) internal pure returns (uint z) {
+        require(y == 0 || (z = x * y) / y == x, 'ds-math-mul-overflow');
+    }
+}
+
+library UniswapV2Library {
```
maxwolff, line 108: are all of these copied from uniswap, or are any ours? ie: which have been audited? would be nice to have a not on where each of these come from. also, maybe call this file `UniswapLib.sol`?

contracts/UniswapOracle.sol
```diff
@@ -0,0 +1,102 @@
...
+pragma experimental ABIEncoderV2;
+
+import "./UniswapContracts.sol";
+
+/**
+ * @notice Uniswap Oracle based on https://github.com/Uniswap/uniswap-v2-periphery/blob/master/contracts/examples/ExampleOracleSimple.sol.
+ * @dev A simple oracle for storing TWAP for 6 assets pairs - ETH, WBTC, ZRX, BAT, REP, DAI.
+ * @author Compound Labs, Inc.
+ */
+contract UniswapOracle {
```
maxwolff, line 11: should we make a `OpenOraclePriceData` (or something) interface, and have the uniswap oracles and `anchorOracle` inherit from it? kind of hard to follow the interfaces here, they dont appear to match in some places

contracts/UniswapContracts.sol
antonina-cherednichenko: Yes, they are just copies. I don't the right way to copy yet, what description to add for that etc.

contracts/UniswapOracle.sol
antonina-cherednichenko: Yes, it should be done. It was not clear how it will be used as an anchor or source. I can definitely adjust now.

contracts/UniswapSlidingWindowOracle.sol
```diff
@@ -0,0 +1,111 @@
...
+            observation.timestamp = block.timestamp;
+            observation.price0Cumulative = price0Cumulative;
+            observation.price1Cumulative = price1Cumulative;
+        }
+    }
+
+    // returns the amount out corresponding to the amount in for a given token using the moving average over the time
+    // range [now - [windowSize, windowSize - periodSize * 2], now]
+    // update must have been called for the bucket corresponding to timestamp `now - windowSize`
+    function getPrice(address pair) external view returns (uint price0Average, uint price1Average) {
```
maxwolff, line 86: hm, so `price0Average` is `reserve1/reserve0`, and `price1Average` is `reserve0/reserve1` [link](https://github.com/Uniswap/uniswap-v2-core/blob/4dd59067c76dea4a0e8e4bfdda41877a6b16dedc/contracts/UniswapV2Pair.sol#L79). we'll have to use price1 or price2 average, depending on which reserve is WETH and which is the token (Uniswap gives it a deterministic but arbitrary [order](https://github.com/Uniswap/uniswap-v2-core/blob/4dd59067c76dea4a0e8e4bfdda41877a6b16dedc/contracts/UniswapV2Factory.sol#L25)). this is something we should probably build into this layer, not have other layers worry about this

tests/UniswapOracleTest.js
```diff
@@ -0,0 +1,292 @@
...
+      expect(price1.price1Average).numEquals(0);
+
+      await sleep(2000);
+      await update("DAI");
+      await sleep(2000);
+      await update("DAI");
+
+      const price2 = await getPrice("DAI");
+      expect(price2.price0Average).toBe("4405985580676962");
+      expect(price2.price1Average).toBe("226963974731472890029");
```
maxwolff, line 152: as mentioned, we only care about one of these prices, right?

tests/UniswapOracleTest.js
```diff
@@ -0,0 +1,292 @@
...
+    });
+
+    it("USDC_ETH pair price", async () => {
+      const price1 = await getPrice("ETH");
+      expect(price1.price0Average).numEquals(0);
+      expect(price1.price1Average).numEquals(0);
+      
+      await sleep(2000);
+      await update("ETH");
+      await sleep(2000);
```
maxwolff, line 116: does this actually work? should we block timestamps instead?

contracts/UniswapSlidingWindowOracle.sol
```diff
@@ -0,0 +1,111 @@
...
+
+    // the desired amount of time over which the moving average should be computed, e.g. 24 hours
+    uint public immutable windowSize;
+    // the number of observations stored for each pair, i.e. how many price observations are stored for the window.
+    // as granularity increases from 1, more frequent updates are needed, but moving averages become more precise.
+    // averages are computed over intervals with sizes in the range:
+    //   [windowSize - (windowSize / granularity) * 2, windowSize]
+    // e.g. if the window size is 24 hours, and the granularity is 24, the oracle will return the average price for
+    //   the period:
+    //   [now - [22 hours, 24 hours], now]
```
maxwolff, line 28: i dont really understand this comment

contracts/UniswapSlidingWindowOracle.sol
maxwolff: actually, itd save us storage to only store the correct price in the `Observation` struct too

tests/UniswapOracleTest.js
antonina-cherednichenko: Yes, but I am adding check for 2, why not :)

contracts/UniswapSlidingWindowOracle.sol
antonina-cherednichenko: Good point. For most of the pairs it's price0, only eth-zrx is price1.

contracts/UniswapSlidingWindowOracle.sol
antonina-cherednichenko: I think it explains this line https://github.com/compound-finance/open-oracle/pull/74/files/3139b366529da7f4a0c20391159090852683b230#diff-3d75c7504e865fedc968a253f45deae0R92
It was inherited from Uniswap example, it should go most likely together with that ^ weird check.

jflatow:
Deprecating PR for #75

PR #95, opened 2020-07-03
Fixes for fixed prices tests: stricter constructor and separate eth

jflatow:


contracts/Uniswap/UniswapAnchoredView.sol
```diff
@@ -79,11 +79,16 @@ contract UniswapAnchoredView is UniswapConfig {
...
-            newObservations[uniswapMarket].acc = cumulativePrice;
+            if (uniswapMarket != address(0)) {
+                require(config.priceSource == PriceSource.REPORTER, "only reported prices utilize an anchor");
+                uint cumulativePrice = currentCumulativePrice(config);
+                oldObservations[uniswapMarket].timestamp = block.timestamp;
+                newObservations[uniswapMarket].timestamp = block.timestamp;
+                oldObservations[uniswapMarket].acc = cumulativePrice;
+                newObservations[uniswapMarket].acc = cumulativePrice;
+            } else {
+                require(config.symbolHash != ethHash, "ETH must have a uniswap market specified");
```
maxwolff, line 90: what about FIXED_USD markets? which have no uniswapMarket but arent ETH?

antonina-cherednichenko, line 90: why only eth ? what about other pairs?

jflatow: good point, not sure why I did that

jflatow: Those don't need a uniswap market right? They would match not eth and would not revert

PR #97, opened 2020-07-04
Remove guards around writing and polish up comments, tests, etc.

jflatow:


contracts/Uniswap/UniswapAnchoredView.sol
```diff
@@ -159,14 +161,10 @@ contract UniswapAnchoredView is UniswapConfig {
             bool withinAnchor = anchorRatio <= upperBoundAnchorRatio && anchorRatio >= lowerBoundAnchorRatio;
 
             if (reporterInvalidated == true) {
-                if (prices[symbolHash] == anchorPrice) {
-                    prices[symbolHash] = anchorPrice;
-                }
+                prices[symbolHash] = anchorPrice;
```
coburncoburn, line 164: probably want priceUpdated here too, right?

PR #99, opened 2020-07-04
Add a uniswap anchored view

jflatow:
This commit adds a Uniswap anchored Open Oracle view. We intend to propose switching the Compound Protocol to use this oracle, together with the Coinbase reporter feed.

The idea is that the price can only be set to a value coming from the reporter, but must be reasonably close to the anchor price. The configuration for the oracle is passed up entirely in the constructor, including which uniswap market to use as an anchor for each reported price.

Tokens can also be configured to have FIXED_ETH or FIXED_USD prices. FIXED_USD prices are constant (since the oracle returns prices denominated in USD). FIXED_ETH prices are a multiple of the ETH price, which must be reported in order to be used.

The view also directly exposes the interface for compatibility with the Compound Protocol, which currently expects to fetch prices by CToken address.

In the unexpected event that the reporter believes their key to be compromised, they may sign a 'rotate' message which can be used to invalidate the reporter. In such an event, the view is configured to fall back to using the anchor price directly.

In order to track the Uniswap price, the price accumulators those markets expose are tracked by block timestamp. Whenever prices are posted, the tracked accumulators are updated in such a way as to always have an accumulator from at least a certain period ago. When a current anchor price is needed, it is always the time-weighted average price over at least the configured anchor period.


contracts/Uniswap/UniswapAnchoredView.sol
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    /**
+     * @dev Fetches the current token/usd price from uniswap, with 6 decimals of precision.
+     */
+    function fetchAnchorPrice(TokenConfig memory config, uint conversionFactor) internal virtual returns (uint) {
+        (uint nowCumulativePrice, uint oldCumulativePrice, uint oldTimestamp) = pokeWindowValues(config);
+
+        // This should be impossible, but better safe than sorry
+        require(block.timestamp > oldTimestamp, "now must come after before");
```
hayesgm, line 207: Can't this happen if this is called twice in the same block for the same asset?

contracts/Uniswap/UniswapAnchoredView.sol
```diff
@@ -0,0 +1,287 @@
...
+        uint timeElapsed = block.timestamp - oldTimestamp;
+
+        // Calculate uniswap time-weighted average price
+        FixedPoint.uq112x112 memory priceAverage = FixedPoint.uq112x112(uint224((nowCumulativePrice - oldCumulativePrice) / timeElapsed));
```
hayesgm, line 211: We should add a comment implying that underflow is desired, or note that it's impossible. Otherwise, this should be done via safe-math (maybe that's implied by the `using FixedPoint`?

contracts/Uniswap/UniswapAnchoredView.sol
```diff
@@ -0,0 +1,297 @@
...
+        uint timeElapsed = block.timestamp - newObservation.timestamp;
+        if (timeElapsed >= anchorPeriod) {
+            emit UniswapWindowUpdate(uniswapMarket, oldObservation.timestamp, newObservation.timestamp, oldObservation.acc, newObservation.acc);
+            oldObservations[uniswapMarket].timestamp = newObservation.timestamp;
+            oldObservations[uniswapMarket].acc = newObservation.acc;
+
+            newObservations[uniswapMarket].timestamp = block.timestamp;
+            newObservations[uniswapMarket].acc = cumulativePrice;
+        }
+        return (cumulativePrice, oldObservation.acc, oldObservation.timestamp);
```
jflatow, line 258: :corn: caught a regression here (fixing), this returns the old `oldObservation` now, not the new one

tests/AnchoredViewTest.js
```diff
@@ -471,7 +471,7 @@ describe('AnchoredView', () => {
     it("sets fallback flag to true if passes", async () => {
       const rotationTarget = '0xAbcdef0123456789000000000000000000000005';
       let encoded = encodeRotationMessage(rotationTarget);
-      // sign rotation message with wrong key
+
```
jflatow, line 474: Why did we get rid of this?

contracts/Uniswap/UniswapAnchoredView.sol
```diff
@@ -0,0 +1,301 @@
...
+    /// @notice The Open Oracle Price Data contract
+    OpenOraclePriceData public immutable priceData;
+
+    /// @notice the number of wei in 1 ETH
+    uint public constant ethBaseUnit = 1e18;
+
+    /// @notice a common scaling factor to maintain precision
+    uint public constant expScale = 1e18;
+
+    /// @notice the Open Oracle Reporter
```
jflatow, line 27: The

contracts/Uniswap/UniswapAnchoredView.sol
```diff
@@ -0,0 +1,301 @@
...
+    uint acc;
+}
+
+contract UniswapAnchoredView is UniswapConfig {
+    using FixedPoint for *;
+
+    /// @notice The Open Oracle Price Data contract
+    OpenOraclePriceData public immutable priceData;
+
+    /// @notice the number of wei in 1 ETH
```
jflatow, line 21: The

contracts/Uniswap/UniswapAnchoredView.sol
```diff
@@ -0,0 +1,301 @@
...
+    uint public constant ethBaseUnit = 1e18;
+
+    /// @notice a common scaling factor to maintain precision
```
jflatow, line 24: A

jflatow:
Merged to master

PR #101, opened 2020-07-08
Quick Window Unit Tests

maxwolff:


contracts/Uniswap/UniswapAnchoredView.sol
```diff
@@ -232,18 +232,18 @@ contract UniswapAnchoredView is UniswapConfig {
         address uniswapMarket = config.uniswapMarket;
         uint cumulativePrice = currentCumulativePrice(config);
 
-        Observation storage newObservation = newObservations[uniswapMarket];
-        Observation storage oldObservation = oldObservations[uniswapMarket];
```
maxwolff, line 236: just load once, then store directly. saves 5k gas. can save another 500 but its kinda ugly, by using storage but then loading each property into memory, but 🤷

jflatow: 🤔 🤦

contracts/Uniswap/UniswapAnchoredView.sol
```diff
@@ -223,18 +223,18 @@ contract UniswapAnchoredView is UniswapConfig {
...
+        Observation memory oldObservation = oldObservations[uniswapMarket];
 
-        // Update new and old observations if elapsed time is bigger or equal to anchor period
+        // Update new and old observations if elapsed time is greater than or equal to anchor period
         uint timeElapsed = block.timestamp - newObservation.timestamp;
         if (timeElapsed >= anchorPeriod) {
-            emit UniswapWindowUpdate(uniswapMarket, oldObservation.timestamp, newObservation.timestamp, oldObservation.acc, newObservation.acc);
-            oldObservation.timestamp = newObservation.timestamp;
-            oldObservation.acc = newObservation.acc;
+            emit UniswapWindowUpdate(uniswapMarket, oldObservation.timestamp, newObservation.timestamp , oldObservation.acc, newObservation.acc);
```
maxwolff, line 232:
```suggestion
            emit UniswapWindowUpdate(uniswapMarket, oldObservation.timestamp, newObservation.timestamp, oldObservation.acc, newObservation.acc);
```

tests/Helpers.js
```diff
@@ -67,5 +73,6 @@ module.exports = {
 	numToHex,
 	uint256,
 	uint,
-	keccak256
+	keccak256,
+  currentBlockTimestamp
```
jflatow, line 77: ,

tests/UniswapAnchoredViewTest.js
```diff
@@ -169,4 +169,56 @@ describe('UniswapAnchoredView', () => {
...
+      expect(newObs2.timestamp).not.numEquals(newObs1.timestamp);
+      expect(oldObs2.acc).numEquals(oldObs1.acc);
+      expect(oldObs2.timestamp).numEquals(oldObs1.timestamp);
+
+      // random new vals
+      await send(mockPair, 'update', ["157323115357569242629001" , "10627310410144389510901", "1592429001", "819360504021542874838907395123146709001", "221435982014902761159721791828816348239901"]);
+      timestamp = Number(await currentBlockTimestamp(web3)) + anchorPeriod;
+      await sendRPC(web3, 'evm_mine', [timestamp]);
+      const tx2 = await postPrices(timestamp, [[['ETH', 201]]], ['ETH']);
+      expect(tx2.events.UniswapWindowUpdate).not.toBe(undefined);
```
jflatow, line 212: Do we ever check fields?

maxwolff: no, ill write one real quick though

contracts/Uniswap/UniswapAnchoredView.sol
```diff
@@ -223,18 +223,18 @@ contract UniswapAnchoredView is UniswapConfig {
...
-            oldObservation.acc = newObservation.acc;
+            oldObservations[uniswapMarket].timestamp = newObservation.timestamp;
+            oldObservations[uniswapMarket].acc = newObservation.acc;
 
-            newObservation.timestamp = block.timestamp;
-            newObservation.acc = cumulativePrice;
+            newObservations[uniswapMarket].timestamp = block.timestamp;
+            newObservations[uniswapMarket].acc = cumulativePrice;
         }
         return (cumulativePrice, oldObservation.acc, oldObservation.timestamp);
```
coburncoburn, line 239: won't this change the return value here in cases where the window markers are moved? before it would have returned the updated value that is eventually in storage under `oldObservations[uniswapMarket]` , now it will return the value that was in oldObservation at the beginning of the function call

jflatow: Yeah this used to return the 'current' old one, now it returns the old one whenever its updated

contracts/Uniswap/UniswapAnchoredView.sol
```diff
@@ -223,18 +223,18 @@ contract UniswapAnchoredView is UniswapConfig {
...
-        Observation storage newObservation = newObservations[uniswapMarket];
-        Observation storage oldObservation = oldObservations[uniswapMarket];
+        Observation memory newObservation = newObservations[uniswapMarket];
+        Observation memory oldObservation = oldObservations[uniswapMarket];
 
-        // Update new and old observations if elapsed time is bigger or equal to anchor period
+        // Update new and old observations if elapsed time is greater than or equal to anchor period
         uint timeElapsed = block.timestamp - newObservation.timestamp;
         if (timeElapsed >= anchorPeriod) {
             emit UniswapWindowUpdate(uniswapMarket, oldObservation.timestamp, newObservation.timestamp, oldObservation.acc, newObservation.acc);
```
coburncoburn, line 232: the event also feels like it should have the new information in it, not the old information. Basically if this read from storage at the end, it would be right, or if it was using memory then V
`
            emit UniswapWindowUpdate(uniswapMarket, newObservation.timestamp, block.timestamp, newObservation.acc, cumulativePrice);
`

PR #102, opened 2020-07-09
Polish: Magic numbers

maxwolff:


contracts/Uniswap/UniswapAnchoredView.sol
```diff
@@ -209,14 +214,14 @@ contract UniswapAnchoredView is UniswapConfig {
...
-        uint anchorPriceUnscaled = mul(priceAverage.decode112with18(), conversionFactor);
+        uint rawUniswapPriceMantissa = mul(priceAverage.decode112with18(), conversionFactor);
         uint anchorPrice;
 
         // Adjust anchor price to val * 1e6 decimals format
         if (config.isUniswapReversed) {
-            anchorPrice = anchorPriceUnscaled / config.baseUnit;
+            anchorPrice = rawUniswapPriceMantissa / config.baseUnit;
         } else {
-            anchorPrice = mul(anchorPriceUnscaled, config.baseUnit) / 1e36;
```
maxwolff, line 219: want to add a comment here too

contracts/Uniswap/UniswapAnchoredView.sol
```diff
@@ -209,14 +217,15 @@ contract UniswapAnchoredView is UniswapConfig {
 
         // Calculate uniswap time-weighted average price
         FixedPoint.uq112x112 memory priceAverage = FixedPoint.uq112x112(uint224((nowCumulativePrice - oldCumulativePrice) / timeElapsed));
-        uint anchorPriceUnscaled = mul(priceAverage.decode112with18(), conversionFactor);
+        uint rawUniswapPriceMantissa = mul(priceAverage.decode112with18(), conversionFactor);
```
hayesgm, line 220: Just to verify, `mantissa` always means `floor(num * 1e18)` encoding?

contracts/Uniswap/UniswapAnchoredView.sol
```diff
@@ -209,14 +217,15 @@ contract UniswapAnchoredView is UniswapConfig {
...
         uint anchorPrice;
 
-        // Adjust anchor price to val * 1e6 decimals format
+        // Adjust rawUniswapPrice according to the units of the non-ETH asset
+        // In the case of ETH, we would have to scale by 1e6 / USDC_UNITS, but since baseUnit2 is 1e6 (USDC), it cancels
         if (config.isUniswapReversed) {
-            anchorPrice = anchorPriceUnscaled / config.baseUnit;
+            anchorPrice = mul(rawUniswapPriceMantissa, ethBaseUnit) / config.baseUnit / expScale;
```
hayesgm, line 226: This is a huge change too. How did this cancel out for markets like WBTC?

contracts/Uniswap/UniswapAnchoredView.sol
```diff
@@ -147,7 +155,6 @@ contract UniswapAnchoredView is UniswapConfig {
             TokenConfig memory config = getTokenConfigBySymbol(symbols[i]);
             string memory symbol = symbols[i];
             bytes32 symbolHash = keccak256(abi.encodePacked(symbol));
-            if (source(messages[i], signatures[i]) != reporter) continue;
```
maxwolff, line 150: calling attention to this bugfix here. since `i` is an iterator for symbol this is and not `signatures[]`, its wrong. we decided we don't need this change anyway, its ok to try to update the view prices even when the messages passed werent signed by the reporter.

contracts/Uniswap/UniswapAnchoredView.sol
maxwolff: not sure what you mean, i mean this to be a non-change. just showing the logic that `expScale` == `ethBaseUnit, though this could just as well be a comment instead of in the code itself, perhaps that would help overflow concerns

contracts/Uniswap/UniswapAnchoredView.sol
maxwolff: yes.. why floor? when does `num * 1e18` cause rounding?

jflatow: I think mantissa means the raw value for an Exponential, which does have 18 decimals precision but probably shouldn't be confused with an amount of wei. Prices are ratios between quantities, and interpreted as Exponentials by the protocol, so mantissa makes sense

contracts/Uniswap/UniswapAnchoredView.sol
jflatow: Yea I agree as a comment would be better

PR #110, opened 2020-07-15
Key observations mappings off symbolHash, add event value tests

maxwolff:
Key `newObservations` `oldObservations` mappings off of symbol hash instead of uniswap, to avoid collisions. also add some more tests for event values

### DIFF
--- a/contracts/Uniswap/UniswapAnchoredView.sol
+++ b/contracts/Uniswap/UniswapAnchoredView.sol
@@ -42,10 +42,11 @@ contract UniswapAnchoredView is UniswapConfig {
     /// @notice Circuit breaker for using anchor price oracle directly, ignoring reporter
     bool public reporterInvalidated;
 
-    /// @notice The old observation for each uniswap market
-    mapping(address => Observation) public oldObservations;
-    /// @notice The new observation for each uniswap market
-    mapping(address => Observation) public newObservations;
+    /// @notice The old observation for each symbolHash
+    mapping(bytes32 => Observation) public oldObservations;
+
+    /// @notice The new observation for each symbolHash
+    mapping(bytes32 => Observation) public newObservations;
 
     /// @notice The event emitted when the stored price is updated
     event PriceUpdated(string symbol, uint price);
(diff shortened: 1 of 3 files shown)


contracts/Uniswap/UniswapAnchoredView.sol
```diff
@@ -57,10 +58,10 @@ contract UniswapAnchoredView is UniswapConfig {
     event ReporterInvalidated(address reporter);
 
     /// @notice The event emitted when the uniswap window changes
-    event UniswapWindowUpdate(address indexed uniswapMarket, uint oldTimestamp, uint newTimestamp, uint oldPrice, uint newPrice);
+    event UniswapWindowUpdate(bytes32 indexed symbolHash, uint oldTimestamp, uint newTimestamp, uint oldPrice, uint newPrice);
 
     /// @notice The event emitted when anchor price is updated
-    event AnchorPriceUpdate(address indexed uniswapMarket, uint anchorPrice, uint nowCumulativePrice, uint oldCumulativePrice, uint oldTimestamp);
+    event AnchorPriceUpdate(bytes32 indexed symbolHash, uint anchorPrice, uint nowCumulativePrice, uint oldCumulativePrice, uint oldTimestamp);
```
maxwolff, line 64: actually, does this event even need cumulative prices? we can always figure out the cumulative prices from `WindowUpdate` events.

antonina-cherednichenko: WindowUpdate is not always emitted

maxwolff: but its emitted on every change. why do we need to know these values if they arent changing?

tests/UniswapAnchoredViewTest.js
```diff
@@ -12,12 +12,13 @@ async function setup(opts)  {
...
-    fixed(1.5e23),
-    fixed(1e22),
-    fixed(1.6e9),//timestamp
-    fixed(8e38),
-    fixed(2e41)// acc
+    "1865335786147",
+    "8202340665419053945756",
+    "1593755855",
+    "119785032308978310142960133641565753500432674230537",
+    "5820053774558372823476814618189",
```
jflatow, line 21: Why change this back?

contracts/Uniswap/UniswapAnchoredView.sol
jflatow: Yeah, maybe not

tests/UniswapAnchoredViewTest.js
maxwolff: was debugging, readded

PR #111, opened 2020-07-15
Open Price Feed Poster Clean-up

hayesgm:
This patch is simply cleaning up some of the Open Price Feed code. I started and I couldn't stop, so it's a nice overhaul just trying to cleanup the functions and interface without significantly changing the behavior. There are some minor CLI differences that we'll need to account for. This is going to be the basis for the branch to handle updating Uniswap for testnets.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2.1
 jobs:
   test_oracle:
     docker:
-      - image: circleci/node:11
+      - image: circleci/node:12
     working_directory: ~/repo
     steps:
       - setup_remote_docker:
@@ -51,7 +51,7 @@ jobs:
 
   test_poster:
     docker:
-      - image: circleci/node:11
+      - image: circleci/node:12
     working_directory: ~/repo/poster
     steps:
       - checkout:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -74,7 +74,7 @@ jobs:
 
   test_reporter_javascript:
     docker:
-      - image: circleci/node:11
+      - image: circleci/node:12
     working_directory: ~/repo/sdk/javascript
     steps:
       - checkout:
--- a/poster/src/poster.ts
+++ b/poster/src/poster.ts
@@ -2,224 +2,173 @@ import { postWithRetries } from './post_with_retries';
 import fetch from 'node-fetch';
 import Web3 from 'web3';
 import { TransactionConfig } from 'web3-core';
-import AbiCoder from 'web3-eth-abi';
-import helpers from './prev_price';
+import {
+  getDataAddress,
+  getPreviousPrice,
+  getSourceAddress
+} from './prev_price';
 import { BigNumber as BN } from 'bignumber.js';
+import { CoinbaseConfig, readCoinbasePayload } from './sources/coinbase';
+import { decodeMessage, encode, zip } from './util';
 
-async function main(sources: string,
-  senderKey: string,
-  viewAddress: string,
-  functionSig: string,
-  gas: number,
-  delta: number,
-  assets: string,
-  web3: Web3) {
-  const payloads = await fetchPayloads(sources.split(","));
-  const filteredPayloads = await filterPayloads(payloads, viewAddress, assets, delta, web3);
-  if (filteredPayloads.length != 0) {
+const GAS_PRICE_API = 'https://api.compound.finance/api/gas_prices/get_gas_price';
+const DEFAULT_GAS_PRICE = 3_000_000_000; // use 3 gwei if api is unreachable for some reason
+
+export async function main(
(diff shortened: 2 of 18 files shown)


poster/src/poster.ts
```diff
@@ -31,195 +42,121 @@ async function main(sources: string,
...
+            },
+            messages: [...acc.messages, payload.messages[index]],
+            signatures: [...acc.signatures, payload.signatures[index]]
+          };
+        } else {
+          return acc;
+        }
+      } else {
+        // Post only prices for supported assets, skip prices for unregistered assets
+        console.debug(`Skipping ${asset} as not part of supported assets: ${JSON.stringify(supportedAssets)}`);
```
coburncoburn, line 90: Nit, may be more clear to filter the assets before or after this reduce action like
```
Object.entries(payload.prices).filter( asset => supportedAssets.includes(asset.toUpperCase())).reduce...
```
loses the logging though 🤷

poster/src/poster.ts
```diff
@@ -31,195 +42,121 @@ async function main(sources: string,
...
-        filteredSignatures.push(payload.signatures[index]);
-      }
-      index++;
+  let filteredPayloads = await Promise.all(payloads.map(async payload => {
+    let sourceAddresses = await Promise.all(payload.messages.map((_, i) => {
+      return getSourceAddress(dataAddress, payload.messages[0], payload.signatures[0], web3);
+    }));
+
+    if ([...new Set(sourceAddresses)].length !== 1) {
+      throw new Error(`Invalid source addresses, got: ${JSON.stringify(sourceAddresses)}`);
```
coburncoburn, line 60: interesting requirement, fine with me, but why did this come up?

poster/src/poster.ts
```diff
@@ -31,195 +42,121 @@ async function main(sources: string,
...
+  const difference = new BN(prevPrice).minus(new BN(price).multipliedBy(1e6)).abs();
+
   return difference.isLessThanOrEqualTo(minDifference);
 }
 
-async function fetchPayloads(sources: string[], fetchFn = fetch): Promise<DelFiReporterPayload[]> {
-  let sourcePromises = sources.map(async (source) => {
+export async function fetchPayloads(sources: string[], fetchFn=fetch): Promise<OpenPriceFeedPayload[]> {
+  let promises = await Promise.allSettled(sources.map(async (sourceRaw) => {
+    let source = sourceRaw.includes('{') ? JSON.parse(sourceRaw) : sourceRaw;
```
coburncoburn, line 116: nice this extra parsing is useful for wrong content-type on response

poster/src/poster.ts
```diff
@@ -2,224 +2,173 @@ import { postWithRetries } from './post_with_retries';
...
+    .forEach(({source, symbol, price, prev}) => {
+      console.log(`Setting Price: source=${source}, symbol=${symbol}, price=${price}, prev_price=${prev}`);
+    });
+
+  return feedItems;
 }
 
 // Checks if new price is less than delta percent different form the old price
-function inDeltaRange(delta: number, price: number, prev_price: number) {
+// Note TODO: price here is uh... a number that needs to be scaled by 1e6?
```
antonina-cherednichenko, line 102: Yes, the price from coinbase api needs to be scaled to 1e6, it's coming in this format:
"BTC": "9212.115", "ETH": "238.03", "XTZ": "3.04975" ...
and prevPrice is scaled to1e6 already.

hayesgm: Yup, but I think the answer is: why is this code written that it accepts two numbers but assumes they are of different scales? I fixed this by descaling the price before this function is called.

poster/src/poster.ts
```diff
@@ -2,224 +2,173 @@ import { postWithRetries } from './post_with_retries';
...
-  assets: string,
-  web3: Web3) {
-  const payloads = await fetchPayloads(sources.split(","));
-  const filteredPayloads = await filterPayloads(payloads, viewAddress, assets, delta, web3);
-  if (filteredPayloads.length != 0) {
+import { CoinbaseConfig, readCoinbasePayload } from './sources/coinbase';
+import { decodeMessage, encode, zip } from './util';
+
+const GAS_PRICE_API = 'https://api.compound.finance/api/gas_prices/get_gas_price';
+const DEFAULT_GAS_PRICE = 3_000_000_000; // use 3 gwei if api is unreachable for some reason
```
maxwolff, line 15: maybe 30 gwei?

.circleci/config.yml
```diff
@@ -51,7 +51,7 @@ jobs:
 
   test_poster:
     docker:
-      - image: circleci/node:11
+      - image: circleci/node:12
```
antonina-cherednichenko, line 54: node:13 here?

PR #112, opened 2020-07-15
Accept the symbol in configuration and store it in a lossless word

jflatow:


jflatow:
Too big of a change at this point, and still too awkward

PR #113, opened 2020-07-16
Test anchoring an erc20 price post

maxwolff:


tests/UniswapAnchoredViewTest.js
```diff
@@ -108,6 +128,18 @@ describe('UniswapAnchoredView', () => {
       expect(await call(priceData, 'getPrice', [reporter.address, 'ETH'])).numEquals(100e6);
     });
 
+    it('should update view if ERC20 price is below anchor bounds', async () => {
```
jflatow, line 131: should not update, right?

tests/UniswapAnchoredViewTest.js
```diff
@@ -121,6 +153,18 @@ describe('UniswapAnchoredView', () => {
...
```
jflatow, line 156: within?

tests/UniswapAnchoredViewTest.js
jflatow: ah, maybe just within?

PR #115, opened 2020-07-16
Last-minute catches

jflatow:


tests/AnchoredViewTest.js
```diff
@@ -471,7 +471,7 @@ describe('AnchoredView', () => {
     it("sets fallback flag to true if passes", async () => {
       const rotationTarget = '0xAbcdef0123456789000000000000000000000005';
       let encoded = encodeRotationMessage(rotationTarget);
-
+      // sign rotation message with wrong key
```
coburncoburn, line 474: is this comment right? looks like it is being signed with the right key

jflatow: Ah, copy-pasta

PR #118, opened 2020-07-17
Add Code Coverage

hayesgm:
This patch adds code coverage to the Open Price Feed system. The coverage will help us make sure we properly cover all cases. I am not currently sure how well this system will work with our immutables, but we'll see. Additionally, this may require extra resources from CircleCI since coverage is significantly more intense (we effectively trace all transactions).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,13 +1,14 @@
 version: 2.1
 
+orbs:
+  codecov: codecov/codecov@1.0.3
+
 jobs:
   test_oracle:
     docker:
       - image: circleci/node:12
     working_directory: ~/repo
     steps:
-      - setup_remote_docker:
-          docker_layer_caching: true
       - run:
         |
           sudo wget https://github.com/ethereum/solidity/releases/download/v0.6.10/solc-static-linux -O /usr/local/bin/solc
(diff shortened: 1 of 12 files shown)


.circleci/config.yml
```diff
@@ -49,6 +52,63 @@ jobs:
       - store_artifacts:
           path: ~/junit-oracle
 
+  test_coverage:
+    docker:
+      - image: circleci/node:11
```
jflatow, line 57: Should we bump node now?

jflatow:
Looks like you rebased and grabbed 2/3 of the commits on `uniswap-polish`, is that intended? If we are going to merge both we probably just want your last commit

.circleci/config.yml
```diff
@@ -41,14 +42,68 @@ jobs:
...
 
+  test_coverage:
+    docker:
+      - image: circleci/node:13
+    working_directory: ~/repo
+    steps:
+      - run:
+        |
+          sudo wget https://github.com/ethereum/solidity/releases/download/v0.6.10/solc-static-linux -O /usr/local/bin/solc
+          sudo chmod +x /usr/local/bin/solc
```
maxwolff, line 59: we already grabbed solc in `test-oracle`, do we have to re download it? does everything get wiped between jobs?

hayesgm: yeah, they run totally separately

hayesgm:
Also FYI, the majority of the real changes occurred here: https://github.com/compound-finance/0x-monorepo/commit/ebceeab03b86468e0cb2b3919482c43413fd88a4

PR #121, opened 2020-07-22
Remove unused contracts and add natspecs

jflatow:


contracts/Uniswap/UniswapAnchoredView.sol
```diff
@@ -83,9 +83,9 @@ contract UniswapAnchoredView is UniswapConfig {
         reporter = reporter_;
         anchorPeriod = anchorPeriod_;
 
-        require(anchorToleranceMantissa_ < 100e16, "anchor tolerance is too high");
-        upperBoundAnchorRatio = 100e16 + anchorToleranceMantissa_;
-        lowerBoundAnchorRatio = 100e16 - anchorToleranceMantissa_;
+        // Allow the tolerance to be whatever the deployer chooses, but prevent under/overflow (and prices from being 0)
+        upperBoundAnchorRatio = anchorToleranceMantissa_ > uint(-1) - 100e16 ? uint(-1) : 100e16 + anchorToleranceMantissa_;
```
hayesgm, line 87: Hm, can this logic be made more obvious? Maybe just use `add(anchorToleranceMantissa_, 100e16)`?

tests/UniswapAnchoredViewTest.js
```diff
@@ -327,13 +327,21 @@ describe('UniswapAnchoredView', () => {
...
       const reporter = web3.eth.accounts.privateKeyToAccount('0x177ee777e72b8c042e05ef41d1db0f17f1fcb0e8150b37cfad6993e4373bdf10');
       const priceData = await deploy('OpenOraclePriceData', []);
-      const anchorMantissa = numToHex(2e18);
-      await expect(
-        deploy('UniswapAnchoredView', [priceData._address, reporter.address, anchorMantissa, 30, []])
-      ).rejects.toRevert("revert anchor tolerance is too high");
+      const anchorPeriod = 30, configs = [];
+      const UINT256_MAX = (1n<<256n) - 1n, exp = (a, b) => BigInt(a) * 10n**BigInt(b);
+
+      const anchorMantissa1 = exp(100, 16);
```
hayesgm, line 336: Cool, I was going to ask for a case for `= 100e16` since it's the corner case. Looks great.

contracts/Uniswap/UniswapAnchoredView.sol
jflatow: But if it overflows you need to also tell it what to cap to (and name that fn), so seems like inlining is more clear imo 🤷‍♂️

PR #124, opened 2020-07-28
Explicitly revert if attempting to post non-reporter prices

jflatow:


contracts/Uniswap/UniswapAnchoredView.sol
```diff
@@ -227,16 +228,17 @@ contract UniswapAnchoredView is UniswapConfig {
         // Calculate uniswap time-weighted average price
         // Underflow is a property of the accumulators: https://uniswap.org/audit.html#orgc9b3190
         FixedPoint.uq112x112 memory priceAverage = FixedPoint.uq112x112(uint224((nowCumulativePrice - oldCumulativePrice) / timeElapsed));
-        uint rawUniswapPriceMantissa = mul(priceAverage.decode112with18(), conversionFactor);
+        uint rawUniswapPriceMantissa = priceAverage.decode112with18();
+        uint unscaledPriceMantissa = mul(rawUniswapPriceMantissa, conversionFactor);
```
hayesgm, line 232: This is just for code clarity?

jflatow: Yeah, its not really the 'raw uniswap price' since the conversion factor is applied

PR #127, opened 2020-07-30
Toni/multiple price deltas

antonina-cherednichenko:
Add different price deltas for assets.

### DIFF
--- a/poster/src/index.ts
+++ b/poster/src/index.ts
@@ -3,6 +3,19 @@ import { main } from './poster';
 import Web3 from 'web3';
 import yargs from 'yargs';
 
+// default price delta triggering an update for the supported asset in %
+const defaultDeltas = {
+  'ETH': 1,
+  'BTC': 1,
+  'DAI': 0.5,
+  'REP': 1.5,
+  'ZRX': 1.5,
+  'BAT': 1.5,
+  'KNC': 1.5,
+  'LINK': 1.5,
+  'COMP': 1.5
+}
+
 async function run() {
   const parsed = yargs
     .env('POSTER')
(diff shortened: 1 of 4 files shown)


poster/src/index.ts
```diff
@@ -14,9 +27,14 @@ async function run() {
     .option('timeout', {alias: 't', description: 'how many seconds to wait before retrying with more gas', type: 'number', default: 180})
     .option('gas-limit', {alias: 'g', description: 'how much gas to send', type: 'number', default: 4000000})
     .option('gas-price', {alias: 'gp', description: 'gas price', type: 'number'})
-    .option('price-delta', {alias: 'd', description: 'the min required difference between new and previous asset price for price update on blockchain', type: 'number', default: 1})
     .option('asset', {alias: 'a', description: 'A list of supported token names for posting prices', type: 'array', default: ['BTC', 'ETH', 'DAI', 'REP', 'ZRX', 'BAT', 'KNC', 'LINK', 'COMP']})
+    // order of `asset` and `price-deltas` should match
+    .option('price-deltas', {
+      alias: 'd', description: 'the min required difference between new and previous asset price for the update on blockchain', type: 'array',
+      default: [defaultDeltas['BTC'], defaultDeltas['ETH'], defaultDeltas['DAI'], defaultDeltas['REP'], defaultDeltas['ZRX'], defaultDeltas['BAT'], defaultDeltas['KNC'], defaultDeltas['LINK'], defaultDeltas['COMP']]
```
coburncoburn, line 34: is it not possible to pass an object here with named values instead of a list? relying on the ordering of these seems kind of error prone

coburncoburn: then we could skip rebuilding the object below

jflatow: +1

antonina-cherednichenko: I wish, but object type  is not supported by yarg arguments.

````type: one of the following strings
'array': synonymous for array: true, see array()
'boolean': synonymous for boolean: true, see boolean()
'count': synonymous for count: true, see count()
'number': synonymous for number: true, see number()
'string': synonymous for string: true, see string()```

Comment (antonina-cherednichenko) [reply poster/src/index.ts:34]: I can do string with json object, just realized that

**Comment by antonina-cherednichenko** [comment]
Replaced arrays with JSON strings for `price-deltas`, `testnet-uniswap-pairs`, `mainnet-uniswap-pairs`, looks much better now.

File: poster/src/index.ts [hunk 2]
```diff
@@ -33,21 +46,30 @@ async function run() {
   const timeout = parsed['timeout'];
   const gas_limit = parsed['gas-limit'];
   const gas_price = parsed['gas-price'];
-  const price_delta = parsed['price-delta'];
+  const price_deltas = JSON.parse(parsed['price-deltas']);
   const assets = <string[]>parsed['asset'];
 
-  // parameters for testnets only
+  // check that price deltas are set up for all assets, otherwise set default delta value
+  assets.forEach(asset => {
```
hayesgm, line 53: FWIW, you could probably just do:
```javascript
const priceDeltasWithDefaults = {
  ...defaultDeltas,
  ...price_deltas
}
```

PR #130, opened 2020-08-04
Add gas estimate calculations and safe processing of `out of gas` error

antonina-cherednichenko:


poster/src/post_with_retries.ts
```diff
@@ -55,6 +59,14 @@ async function postWithRetries(transaction: TransactionConfig, signerKey: string
       };
     }
 
+    if (maybeIsOutOfGas(e)) {
```
antonina-cherednichenko, line 62: I added a gas increase ability for the situation when a gas estimate is not very accurate

jflatow:
Looks great

poster/src/post_with_retries.ts
```diff
@@ -55,6 +59,14 @@ async function postWithRetries(transaction: TransactionConfig, signerKey: string
...
+      console.log('Transaction is possibly out of gas, increasing gas limit');
+      transaction = {
+        ...transaction,
+        gas: Math.floor(Number(transaction.gas) * GAS_ADJUSTMENT)
```
jflatow, line 66: Maybe we should always apply at least one level of adjustment?

antonina-cherednichenko: @coburncoburn @hayesgm any thoughts here? I can increase the gas limit on each retry to simplify logic here

PR #131, opened 2020-08-05
Increase initial gas estimate by 50%

antonina-cherednichenko:
Adjust initial gas estimate by 1.5

### DIFF
--- a/poster/src/poster.ts
+++ b/poster/src/poster.ts
@@ -46,11 +46,13 @@ export async function main(
 
     const trxData = buildTrxData(feedItems, functionSig);
     const gasEstimate = await web3.eth.estimateGas({data: trxData, to: viewAddress});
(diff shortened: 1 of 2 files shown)


poster/src/poster.ts
```diff
@@ -46,11 +46,13 @@ export async function main(
 
     const trxData = buildTrxData(feedItems, functionSig);
     const gasEstimate = await web3.eth.estimateGas({data: trxData, to: viewAddress});
+    // Make gas estimate safer by 50% adjustment
+    const gastEstimateAdjusted = Math.floor(gasEstimate * 1.5);
```
jflatow, line 50: I guess there are two `GAS_ADJUSTMENT`s

antonina-cherednichenko: is that okay? I thought that keeping adjustment inside the `catch` clause if posting fails is useful

PR #133, opened 2020-08-13
delete default price deltas

toninorair:
Delete default price deltas

### DIFF
--- a/poster/src/index.ts
+++ b/poster/src/index.ts
@@ -38,7 +38,7 @@ async function run() {
 
   // check that price deltas are set up for all assets
   assets.forEach(asset => {
-    if (!price_deltas[asset]) {
+    if (price_deltas[asset] == undefined) {
       throw new TypeError(`For each asset price delta should be provided, ${asset} asset is not properly configured`)
     }
   });
--- a/poster/src/index.ts
+++ b/poster/src/index.ts
@@ -3,19 +3,6 @@ import { main } from './poster';
 import Web3 from 'web3';
 import yargs from 'yargs';
 
-// default price delta triggering an update for the supported asset in %
-const defaultDeltas = {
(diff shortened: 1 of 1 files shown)


poster/src/index.ts
```diff
@@ -49,10 +36,10 @@ async function run() {
   const price_deltas = JSON.parse(parsed['price-deltas']);
   const assets = <string[]>parsed['asset'];
 
-  // check that price deltas are set up for all assets, otherwise set default delta value
+  // check that price deltas are set up for all assets
   assets.forEach(asset => {
     if (price_deltas[asset] == undefined) {
-      price_deltas[asset] = defaultDeltas[asset];
+      throw new TypeError(`For each asset price delta should be provided, ${asset} asset is not properly configured`)
```
hayesgm, line 42: Only question is: maybe we just want to have a sane default to make life easy, like 5%? Not that we're using it, but it's easier than getting yelled at. Thoughts?

hayesgm (approved):
Looks great- one comment about maybe having _some_ default

zack-eth (approved):
LGTM, here is the corresponding PR: https://github.com/compound-finance/open-price-feed-poster/pull/12

poster/src/index.ts
antonina-cherednichenko: that's what we had before, but then I switched it to different deltas. I would say let's just set all deltas in the poster, it's clear approach and it forces us to properly configure poster and really understand deltas

antonina-cherednichenko: kind of 'fail early, fail hard' if you forget about an asset

hayesgm: 👍 Sounds good. I like this more now come to think of it. Explicit rocks.

PR #138, opened 2020-09-08
Add Open Oracle v1

hayesgm:
This patch adds an open oracle v1 which is meant to align with the interface of the v1 oracle but support reading data from the open oracle. We simply use the token configuration information to pull the correct prices, in terms of Eth, for the given asset requeest.

### DIFF
--- a/contracts/v1/OpenOracleV1.sol
+++ b/contracts/v1/OpenOracleV1.sol
@@ -1,0 +1,44 @@
+// SPDX-License-Identifier: GPL-3.0
+
+pragma solidity ^0.6.10;
+pragma experimental ABIEncoderV2;
+
+import "../Uniswap/UniswapAnchoredView.sol";
+
+interface IPriceOracle {
+	function assetPrices(address) external returns (uint256);
+	function getPrices(address) external returns (uint256);
+}
+
+/**
+ * @title Open Oracle Adapater for v1 Interface
+ * @author Compound Labs, Inc.
+ */
+contract OpenOracleV1 is IPriceOracle {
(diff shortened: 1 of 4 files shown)


contracts/v1/OpenOracleV1.sol
```diff
@@ -0,0 +1,44 @@
...
+ */
+contract OpenOracleV1 is IPriceOracle {
+	UniswapAnchoredView public immutable openOracleView;
+
+	constructor(UniswapAnchoredView openOracleView_) public {
+		openOracleView = openOracleView_;
+	}
+
+	function getPrices(address asset) public override returns (uint256) {
+		UniswapConfig.TokenConfig memory ethConfig = openOracleView.getTokenConfigBySymbol("ETH");
```
coburncoburn, line 25: holy crow, good thing those get configs were public

jflatow (approved):
💪 👏

contracts/v1/OpenOracleV1.sol
hayesgm: Yeah, otherwise this would have been way more complex to build

coburncoburn (approved):
looks worth deploying to me, followed by some sanity checkin'

melyalex (approved):

Asim-tech235:
> This patch adds an open oracle v1 which is meant to align with the interface of the v1 oracle but support reading data from the open oracle. We simply use the token configuration information to pull the correct prices, in terms of Eth, for the given asset requeest.

Thugnasty777 (approved):

philipjonsen (approved):

PR #139, opened 2020-09-10
Upgrade saddle to version with json fix

antonina-cherednichenko:
Upgrade saddle to the version with JSON number fix


hayesgm (approved):
lgtm

PR #140, opened 2020-10-06
Update and rename sdk/javascript/examples/fixed.js to sdk/javascript/…

kurczab11:
…przykłady/fixed.js


jflatow:
I'm not sure why we would change the directory name from English `examples` to Polish?

PR #141, opened 2020-10-08
Revert "Upgrade saddle to version with json fix"

kurczab11:
Reverts compound-finance/open-oracle#139


jflatow:
We don't want to move the saddle version backwards :) If you have a reason for this, please explain, but otherwise we may start treating these PRs like spam

PR #143, opened 2020-11-02
price range proofs

AyushKaul:


jflatow:
Wow interesting! Not sure I follow it though - is the proof that the reporter signed a price between min and max, or is the reporter obsolete here? How is the proof generated?

AyushKaul:
the proof verifies that the reporter signed a price between min and max. the proof is generated using zokrates, it has to be verified before posting.

PR #147, opened 2021-01-07
Fix the potential anchor price calculation bug.

RickieRao:
Remove the config.isUniswapReversed flag check for anchor price calculation as the scale will shift away for tokens that fit:
1. Address larger than WETH address
2. Decimal NOT equal to WETH decimal(18)


hayesgm (approved):

Asim-tech235 (commented):
☹️

Asim-tech235 (commented):
- [ ] #> ~~``~~

